Issue #17, opened 2020-01-19
Open Example Project only copies one level deep (VSC-220)


Devenda:
**Describe the bug**
Creating an example only copies one level of folders.

**To Reproduce**
Open the example menu, choose a project containing multiple levels of folders (eg restful server, touch_pad_read,...) click "Create project using example...".
results in nested folders being treated as files:
![afbeelding](https://user-images.githubusercontent.com/2527506/72682239-375f8380-3acb-11ea-8dcf-ff32b072eaaa.png)

**Expected behavior**
Fully recursively copied folders and files.

**Screenshots**
Leftside: original example in esp-idf folder, rightside: the folder after copy.
![afbeelding](https://user-images.githubusercontent.com/2527506/72682196-d3d55600-3aca-11ea-99f8-94cb6cecb0e6.png)

**Environment:**
plugin version 0.1.2
Windows 10

Devenda:
Additionally files in the project root are not copied either:
![afbeelding](https://user-images.githubusercontent.com/2527506/72682323-16e3f900-3acc-11ea-8180-229e3fe51788.png)

brianignacio5:
Thanks for reporting this issue!

Will fix soon!

PR #18, opened 2020-01-20
Fix/examples copy (VSC-221)

brianignacio5:
Closes https://github.com/espressif/vscode-esp-idf-extension/issues/17

* Fix examples copy error in https://github.com/espressif/vscode-esp-idf-extension/issues/17
* Add extension configuration explanation
* Updates template tasks.json with OPENOCD_SCRIPTS from extension `idf.customExtraVars` with fallback in system env OPENOCD_SCRIPTS.

### DIFF
--- a/src/downloadManager.ts
+++ b/src/downloadManager.ts
@@ -15,8 +15,8 @@
 import * as del from "del";
 import { https } from "follow-redirects";
 import * as fs from "fs";
+import { ensureDir, pathExists } from "fs-extra";
 import * as http from "http";
-import * as mkdirp from "mkdirp";
 import * as path from "path";
 import * as url from "url";
 import * as vscode from "vscode";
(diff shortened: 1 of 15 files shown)


src/utils.ts
```diff
@@ -177,42 +177,39 @@ export function createSkeleton(curWorkspacePath: string, chosenTemplateDir: stri
...
-                Logger.errorNotify(err.message, err);
-                return;
-            }
-            files.forEach((file) => {
-                copyTarget(path.join(curDir, file), path.join(curDestDir, file));
-            });
-        });
-    });
-
+export function copyDirectory(srcDirPath: string, destDirPath: string) {
```
SoumeshBanerjee: Maybe use `fs-extra` or some other lib for copying dir.

src/downloadManager.ts
```diff
@@ -81,7 +81,7 @@ export class DownloadManager {
         const destPath = this.getToolPackagesPath(["dist"]);
         const absolutePath: string = this.getToolPackagesPath(["dist", fileName]);
 
-        await utils.checkFileExists(absolutePath).then(async (pkgExists) => {
+        await pathExists(absolutePath).then(async (pkgExists) => {
```
SoumeshBanerjee: maybe remove `await` since we are not using the result return and using `.then()` to process the result instead

brianignacio5: But we need for the process to wait to finish or else the install part will just execute (on Windows has happened on Mac it doesn't) before this part being finished.

PR #22, opened 2020-01-23
Github Actions for Releases

SoumeshBanerjee:
Add workflow to create automated release when there are tagged commits pushed.
- Release to the VSCode Marketplace
- Release to the Github Releases and Upload the `.vsix` generated

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,47 @@
+name: Release
+
+on:
+  push:
+    tags: 
+      - "v*"
+
+jobs:
+  build:
+    name: Upload .vxis in release asset
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


SoumeshBanerjee:
> Inspires me to to do the same for IDF releases
@projectgus  IMO, Most challenging thing to do would be to add `CHANGELOG` intelligently and then add those to the release body.

projectgus:
> IMO, Most challenging thing to do would be to add CHANGELOG intelligently and then add those to the release body.

@SoumeshBanerjee  Agree. But as a first cut it looks like we can create a draft release, and upload the .zip file to the release and also to dl.espressif.com. That cuts out of a lot of the "busywork" in the release process, just need to paste in the Release Notes at that point.

PR #43, opened 2020-02-18
read Cmakelists file for project name (VSC-248)

brianignacio5:
Instead of using idf.projectName user configuration, read Cmakelists.txt file to resolve project name.

VSC-230


src/workspaceConfig.ts
```diff
@@ -45,16 +44,27 @@ export function updateIdfComponentsTree(projectDescriptionPath: string) {
...
+export function getProjectName(workspace: string): Promise<string> {
+    const cmakeListFile = path.join(workspace, "CMakeLists.txt");
+    return new Promise((resolve, reject) => {
+        try {
+            if (!utils.fileExists(cmakeListFile)) {
+                reject(`${cmakeListFile} doesn't exist.`);
+            }
+            fs.readFile(cmakeListFile, (err, data) => {
+                if (err) {
+                    Logger.errorNotify(err.message, err);
```
SoumeshBanerjee: maybe use `Logger.error()` instead of `errorNotify` since this is an utility function, and caller should decide how to display the error

src/workspaceConfig.ts
```diff
@@ -45,16 +44,27 @@ export function updateIdfComponentsTree(projectDescriptionPath: string) {
...
-    fs.readFile(projectDescriptionPath, (err, data) => {
-        if (err) {
-            Logger.errorNotify(err.message, err);
-            return;
+export function getProjectName(workspace: string): Promise<string> {
+    const cmakeListFile = path.join(workspace, "CMakeLists.txt");
+    return new Promise((resolve, reject) => {
+        try {
+            if (!utils.fileExists(cmakeListFile)) {
+                reject(`${cmakeListFile} doesn't exist.`);
```
SoumeshBanerjee: reject with error type not string

src/workspaceConfig.ts
```diff
@@ -45,16 +44,27 @@ export function updateIdfComponentsTree(projectDescriptionPath: string) {
...
+            fs.readFile(cmakeListFile, (err, data) => {
+                if (err) {
+                    Logger.errorNotify(err.message, err);
+                    reject(err);
+                }
+                const content = data.toString();
+                const projectMatches = content.match(/(?:project\()(.*?)(?:\))/);
+                if (projectMatches && projectMatches.length > 0) {
+                    resolve(projectMatches[1]);
+                }
```
SoumeshBanerjee: else if missing !

if the condition is not satisfies the promise will not be rejected, neither be resolved, leading to an indeterministic state

src/extension.ts
```diff
@@ -357,6 +356,11 @@ export async function activate(context: vscode.ExtensionContext) {
 
     const debugProvider = new IdfDebugConfigurationProvider();
     context.subscriptions.push(vscode.debug.registerDebugConfigurationProvider("cppdbg", debugProvider));
+
+    registerIDFCommand("espIdf.getProjectName", () => {
```
SoumeshBanerjee: do we need to guard this command since it requires an open workspace to work?

src/extension.ts
```diff
@@ -357,6 +356,11 @@ export async function activate(context: vscode.ExtensionContext) {
...
+        return getProjectName(workspaceRoot.fsPath);
```
SoumeshBanerjee: maybe we need to await here, since this returns a promise and not the project name, its always a good practice to return concrete value or display error to the user here, when the command is executed

src/workspaceConfig.ts
```diff
@@ -45,16 +44,27 @@ export function updateIdfComponentsTree(projectDescriptionPath: string) {
...
+            if (!utils.fileExists(cmakeListFile)) {
+                reject(`${cmakeListFile} doesn't exist.`);
+            }
+            fs.readFile(cmakeListFile, (err, data) => {
+                if (err) {
+                    Logger.errorNotify(err.message, err);
+                    reject(err);
+                }
+                const content = data.toString();
+                const projectMatches = content.match(/(?:project\()(.*?)(?:\))/);
```
igrr: Ouch, this seems to be very fragile... if the argument to the project function is a variable reference, then this won't work. Any chance we can get the project name from the project metadata file instead (build/project_description.json, "project_name" key). The project metadata file becomes available after CMake configuration phase runs.

SoumeshBanerjee: Also, If we use `build/project_description.json` then we can do `vscode.workspace.createFileSystemWatcher` for watching the `CMakeLists.txt` and run `CMake configuration phase` on change and update the project name, if we want to add kind of instant update

igrr: Yes, i think this is what VS Code CMake extension does — it watches CMakeLists.txt files and re-runs CMake when the file changes.

If you do end up parsing the project description, you can get more things from it. For instance, the list of project components, the path of the sdkconfig file, toolchain prefix, monitor baud rate, and so on.

brianignacio5: Thanks for the feedback, will add them soon.

PR #50, opened 2020-03-02
configuration save destination (VSC-255)

### LINKED ISSUE #46: Option to select where settings should be saved (VSC-251)
Currently, Onboarding works by appending the settings to the user settings, and building the project won't work if those settings are not specifically there.

There should be a way to choose if the settings should be user-wide and workspace-wide. An option for Onboarding and another one for the Command Palette. Consequently, the extension would default to look for workspace settings first before moving on to look at user settings, following the regular VSCode flow of workspace settings overriding user's, or another option to choose which one should take precedence, should there ever be a need for such a thing.

The main reason why this is a must have it's because it makes it easy to work with multiple projects, versions of the toolchain, and versions of ESP-IDF. Another reason is pollution of user configurations, especially if VSCode isn't used solely for ESP32 development. I'm sure there are others, but these are the ones that I can think of right now.

brianignacio5:
Fix #46 

Enable the user to select where to save configuration settings.

On vscode commands -> extra step to define where to save configuration settings.

On Onboarding -> dropdown on first window to select where to save configuration settings for the whole onboarding experience.

### DIFF
--- a/src/espIdf/serial/serialPort.ts
+++ b/src/espIdf/serial/serialPort.ts
@@ -20,7 +20,7 @@ import * as vscode from "vscode";
 import * as idfConf from "../../idfConfiguration";
 import { LocDictionary } from "../../localizationDictionary";
 import { Logger } from "../../logger/logger";
-import { PreCheck, spawn } from "../../utils";
+import { spawn } from "../../utils";
 import { SerialPortDetails } from "./serialPortDetails";
 
 export class SerialPort {
@@ -71,12 +71,14 @@ export class SerialPort {
   }
 
   private async updatePortListStatus(l: string) {
-    await idfConf.writeParameter("idf.port", l);
-    const portHasBeenSelectedMsg = this.locDic.localize(
-      "serial.portHasBeenSelectedMessage",
-      "Port has been updated to "
-    );
-    Logger.infoNotify(portHasBeenSelectedMsg + l);
+    await idfConf.chooseConfigurationTarget().then(async (target) => {
+      await idfConf.writeParameter("idf.port", l, target);
+      const portHasBeenSelectedMsg = this.locDic.localize(
+        "serial.portHasBeenSelectedMessage",
+        "Port has been updated to "
+      );
+      Logger.infoNotify(portHasBeenSelectedMsg + l);
+    });
   }
 
   private list(): Thenable<SerialPortDetails[]> {
--- a/src/espIdf/serial/serialPort.ts
+++ b/src/espIdf/serial/serialPort.ts
@@ -71,14 +71,13 @@ export class SerialPort {
   }
 
   private async updatePortListStatus(l: string) {
-    await idfConf.chooseConfigurationTarget().then(async (target) => {
-      await idfConf.writeParameter("idf.port", l, target);
-      const portHasBeenSelectedMsg = this.locDic.localize(
-        "serial.portHasBeenSelectedMessage",
-        "Port has been updated to "
-      );
-      Logger.infoNotify(portHasBeenSelectedMsg + l);
-    });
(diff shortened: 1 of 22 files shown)


src/extension.ts
```diff
@@ -261,64 +261,62 @@ export async function activate(context: vscode.ExtensionContext) {
...
-                }
-                if (msg && paramName) {
-                    currentValue = idfConf.readParameter(paramName);
-                    if (currentValue instanceof Array) {
-                        currentValue = currentValue.join(",");
-                    }
-                    idfConf.updateConfParameter(paramName, msg, currentValue, option.label);
+        vscode.window.showQuickPick(
+            [
+                { description: "Device target (esp32, esp32s2beta)",
```
kumekay: May be we should n't hardcode list of targets here? Recent master supports esp32s2, while esp32s2beta is only supported by 4.1 branch?

src/extension.ts
```diff
@@ -412,17 +390,21 @@ export async function activate(context: vscode.ExtensionContext) {
...
                 { placeHolder: enterDeviceTargetMsg },
-            ).then((selected) => {
+            ).then(async (selected) => {
                 if (typeof selected === "undefined") {
                     return;
                 }
-                idfConf.writeParameter("idf.adapterTargetName", selected.target);
+                const configurationTarget = await idfConf.chooseConfigurationTarget();
+                idfConf.writeParameter("idf.adapterTargetName", selected.target, configurationTarget);
                 if (selected.target === "esp32") {
```
kumekay: Is it possible to support new targets without hardcoding them here? 
I.e. move to some config-file, or at least separate JS file.
(and we should add esp32s2, probably in separate PR)

src/extension.ts
brianignacio5: Is there a way to get the supported targets from esp-idf ? 
Will move to a separated supported target config file.

kumekay: There is `idf.py --list-targets` command, that is supported starting v4.1

kumekay: Not sure, that this way is good enough to be used by IDE

brianignacio5: We still need to have some sort of target file to map it with related openOCD config files as we do here. We could have all targets in extension and show error to user if they select a non supported target for the current idf.espIdfPath, what do you think?

kumekay: I cannot come up with a better solution, so this is ok

src/idfConfiguration.ts
```diff
@@ -51,14 +56,67 @@ export function readParameter(param: string) {
   return paramValue;
 }
 
-export async function writeParameter(param: string, newValue) {
+export function chooseConfigurationTarget() {
```
SoumeshBanerjee: Maybe we need to persist the user preference, because its tedious for user to select everytime!

src/espIdf/serial/serialPort.ts
```diff
@@ -71,12 +71,14 @@ export class SerialPort {
...
 
   private async updatePortListStatus(l: string) {
-    await idfConf.writeParameter("idf.port", l);
-    const portHasBeenSelectedMsg = this.locDic.localize(
-      "serial.portHasBeenSelectedMessage",
-      "Port has been updated to "
-    );
-    Logger.infoNotify(portHasBeenSelectedMsg + l);
+    await idfConf.chooseConfigurationTarget().then(async (target) => {
+      await idfConf.writeParameter("idf.port", l, target);
```
SoumeshBanerjee: lets just use `async/await` and not mixin with `.then/.catch` block

The above code can be re-written like this which makes it easier to read and does the same thing

```js
const target = await idfConf.chooseConfigurationTarget();
await idfConf.writeParameter("idf.port", l, target);
```

PR #51, opened 2020-03-03
add gif in readme (VSC-256)

brianignacio5:
Add gif with extension current features


kumekay:
Brian, for me the gif blinking too fast (~.2 sec per frame), rendering it almost useless, is it the same for you? May be we can slow down it a bit (and remove some frames)

PR #55, opened 2020-03-10
Heap Tracing Feature (VSC-259)

pwmb:
Initial implementation for Heap Tracing.

Provide basic functionality for the end user to perform heap tracing with minimum setup and clicks.

It uses the TCL connection to communicate to the openOCD over TCP, once the tracing is done the report will be available in the report archives for viewing for the user.

Once a report is requested to be viewed by the user, it will present with a horizontal timeline of allocation, deallocation and logs.

Example ESP-IDF Application for testing heap tracing [can be obtained here](https://github.com/espressif/esp-idf/tree/master/examples/system/sysview_tracing_heap_log)


src/espIdf/tracing/heapTraceManager.ts
```diff
@@ -0,0 +1,285 @@
...
+                        },
+                    })
+                    .buildCommand({
+                        command: "reg pc", responseHandler: {
+                            type: "commandResponse",
+                            handler: [addresses.heap_trace_start],
+                        },
+                    })
+                    .buildCommand({
+                        command: `rbp 0x${addresses.heap_trace_start}`,
```
igrr: Question, why do we need to set breakpoints on the current PC before/after the sysview command? (cc @gerekon)

src/espIdf/openOcd/openOcdManager.ts
```diff
@@ -95,6 +98,24 @@ export class OpenOCDManager extends EventEmitter {
...
+        const tclClient = new TCLClient(this.tclConnectionParams);
+        if (!await tclClient.isOpenOCDServerRunning()) {
+            const resp = await vscode.window.showInformationMessage("OpenOCD is not running, do you want to launch it?",
+                { modal: true },
+                { title: "Yes" },
+                { title: "Cancel", isCloseAffordance: true },
+            );
+            if (resp && resp.title === "Yes") {
+                await OpenOCDManager.init().start();
+                await sleep(1000);
```
igrr: Perhaps this delay can be handled in `isOpenOCDServerRunning`?

src/utils.ts
```diff
@@ -338,13 +339,42 @@ export function isStringNotEmpty(str: string) {
...
+    }
+    const elfFiles = [];
+    fs.readdirSync(elfPath).forEach((file) => {
+        if (file.endsWith(".elf")) {
+            elfFiles.push({ label: file, description: path.join(elfPath, file) });
+        }
+    });
+    if (elfFiles.length > 1) {
+        const pickedElf = await vscode.window.showQuickPick(elfFiles, {
+            placeHolder: "Select ELF File to be use for the report generation",
```
igrr: Instead of prompting the user, can we get the ELF file name from the project metadata file?

SoumeshBanerjee: Yes! that was the initial code, but changed it in case user have multiple elf file in future like build for various arch or `esp32`/ `esp32s2`
Also this would prompt user to select elf file only when there are multiple `.elf` files

igrr: We can't have multiple ELF files for different architectures in the build directory — the build system will refuse to build for a different target without running `fullclean`. So I think we can do this automatically.

src/espIdf/tracing/tools/xtensa/addr2line.ts
```diff
@@ -0,0 +1,44 @@
...
+import { constants } from "fs";
+import { canAccessFile } from "../../../../utils";
+import { XtensaTools } from "./abstractXtensaTools";
+
+export class Addr2Line extends XtensaTools {
+    private readonly elfFilePath: string;
+    private readonly address: string;
+
+    constructor(workspaceRoot: vscode.Uri, elfFilePath: string, address: string) {
+        super(workspaceRoot, "xtensa-esp32-elf-addr2line");
```
igrr: Toolchain prefix depends on the target

src/espIdf/tracing/tools/xtensa/nm.ts
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+import * as vscode from "vscode";
+
+import { constants } from "fs";
+import { canAccessFile } from "../../../../utils";
+import { XtensaTools } from "./abstractXtensaTools";
+
+export class Nm extends XtensaTools {
```
brianignacio5: Maybe use a more descriptive name here?

SoumeshBanerjee: `XtensaTools` is an abstract tools class which abstracts some xtensa tools and `Nm` is the name of one of such tool

src/views/tracing/index.html
```diff
@@ -0,0 +1,101 @@
...
+                v-show="heapView.stats"
+                v-bind:persistent-count="persistentCount"
+                v-bind:persistent-bytes="persistentBytes"
+                v-bind:transient-count="transientCount"
+                v-bind:total-bytes="totalBytes"
+                v-bind:total-count="totalCount"
+              />
+            </div>
+
+            <quick-action-menu v-on:change="heapViewChange"></quick-action-menu>
```
brianignacio5: Maybe move a level up the quick action menu component? When window size is small, buttons are shown at the button of the scrollable window.

SoumeshBanerjee: Yes... will stick it at the top

src/espIdf/tracing/heapTraceManager.ts
gerekon: @igrr  This command removes breakpoint.

SoumeshBanerjee:
@gerekon 

> It seems that this PR includes much more than just heap tracing stuff

Yes but those changes are just related to formatting / linting of code and will not be part of file changed once #59 is merged

src/espIdf/tracing/appTraceManager.ts
```diff
@@ -0,0 +1,276 @@
...
+          join(workspacePath, "trace", `trace_${new Date().getTime()}.trace`)
+        );
+        const pollPeriod = idfConf.readParameter("trace.poll_period");
+        const traceSize = idfConf.readParameter("trace.trace_size");
+        const stopTmo = idfConf.readParameter("trace.stop_tmo");
+        const wait4halt = idfConf.readParameter("trace.wait4halt");
+        const skipSize = idfConf.readParameter("trace.skip_size");
+        const startTrackingHandler = this.sendCommandToTCLSession(
+          [
+            "esp32",
```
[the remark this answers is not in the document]
SoumeshBanerjee: @gerekon Fixed all the prefix with `esp` instead of __target name__

pwmb:
@brianignacio5 Merging this to avoid daily sync and merge conflict resolution with master, will release `v0.3.0` once specific version of `OpenOCD` and tracing tools with `json` support is released to `esp-idf`

PR #56, opened 2020-03-10
IDF and Tools Metadata, New Project Wizard (VSC-260)

brianignacio5:
This PR introduces the use of metadata.json,  a file generated on user extension folder recording all previous onboarding install.

Features:

1) Add a `Use previous settings` window in `ESP-IDF: Configure ESP-IDF extension`. This window will show a dropdown with all ESP-IDF versions previous installed with this onboarding window. When you select a version, the python virtualenv binary path dropdown is shown as well and ESP-IDF tools paths as input boxes where you can modify the path for each tool separately. A `Check tools are valid` button will use version validation (same as done in Skip tools download step) to validate each path and a error message is shown when a path is invalid. If all path are valid, a `Save settings` button will appear which will save `idf.espIdfPath`, `idf.pythonBinPath` and `idf.customExtraPaths` and `idf.customExtraVars` respectively.

2) Check tools are valid before build for current `idf.espIdfPath` showing error in log file and notification for the missing/invalid tool.

3) Show a message to open  `ESP-IDF: Configure ESP-IDF extension` when `idf.espIdfPath` changes

**Note**: Should we also provide modify previous settings ?


brianignacio5:
Can't use a version field for ESP-IDF paths since it is not static. Best we can do is shown the current version for selected ESP-IDF Path.

src/idfToolsManager.ts
```diff
@@ -86,14 +87,14 @@ export class IdfToolsManager {
...
-    return this.getPackageList(onReqPkgs).then(async (packages) => {
+  public async verifyPackages(pathsToVerify: string, onReqPkgs?: string[]) {
+    return await this.getPackageList(onReqPkgs).then(async (packages) => {
       const promiseArr = {};
       const names = packages.map((pkg) => pkg.name);
       const promises = packages.map((pkg) =>
         this.checkBinariesVersion(pkg, pathsToVerify)
       );
-      return Promise.all(promises).then((versionExistsArr) => {
+      return await Promise.all(promises).then((versionExistsArr) => {
```
SoumeshBanerjee, line 97: `await` `.then` mixin here its better to use just `await` and return that

src/onboarding/espIdfDownload.ts
```diff
@@ -360,3 +363,39 @@ export async function getEspIdfVersions(extensionPath: string) {
...
+    utils.extensionContext.extensionPath,
+    "metadata.json"
+  );
+  const idfMetadata: IPath = {
+    id: uuidv4(),
+    path: idfPath,
+  } as IPath;
+  await utils.doesPathExists(metadataFile).then(async (doesFileExists) => {
+    if (doesFileExists) {
+      await utils
```
SoumeshBanerjee, line 378: here as well `await` `.then` mixin

src/utils.ts
```diff
@@ -195,6 +199,19 @@ export function readFileSync(filePath) {
...
+export function doesPathExists(inputPath: string) {
+  return pathExists(inputPath);
+}
+export function readJson(jsonPath: string) {
+  return readJSON(jsonPath);
+}
+
+export function writeJson(jsonPath: string, object: any) {
+  return writeJSON(jsonPath, object, {
+    spaces: vscode.workspace.getConfiguration().get("editor.tabSize"),
```
SoumeshBanerjee, line 211: `get<T>(section: string): T | undefined;` this could return undefined, maybe we need to provide a default value here

src/utils.ts
```diff
@@ -603,3 +637,137 @@ export async function isBinInPath(binaryName: string, workDirectory: string) {
...
 }
+
+export async function loadMetadata() {
+  const metadataFile = path.join(
+    extensionContext.extensionPath,
+    "metadata.json"
+  );
+  return await doesPathExists(metadataFile).then(async (doesMetadataExist) => {
+    if (doesMetadataExist) {
+      return await readJson(metadataFile).then((metadata: IMetadataFile) => {
```
SoumeshBanerjee, line 648: here too

src/utils.ts
```diff
@@ -603,3 +637,137 @@ export async function isBinInPath(binaryName: string, workDirectory: string) {
...
+}
+
+export async function writeToMetadataFile(toolsInfo: ITool[]) {
+  const metadataFile = path.join(
+    extensionContext.extensionPath,
+    "metadata.json"
+  );
+  await doesPathExists(metadataFile).then(async (doesFileExists) => {
+    if (doesFileExists) {
+      await readJson(metadataFile).then(async (metadata: IMetadataFile) => {
```
SoumeshBanerjee, line 685: here too

src/views/onboarding/UsePrevious.vue
```diff
@@ -0,0 +1,178 @@
...
+                <select v-model="selectedIdfMetadata" id="idf-version-select" @change="changeSelectedIdfPath">
+                    <option v-for="ver in idfVersionsMetadata" :key="ver.id" :value="ver">
+                        {{ ver.path }}
+                    </option>
+                </select>
+                <p>Selected ESP-IDF Path version: {{ previousIdfVersion }}</p>
+                <div v-if="venvVersionsMetadata && venvVersionsMetadata.length > 0">
+                    <label for="python-version-select">Python version:</label>
+                    <br><br>
+                    <select v-model="selectedVenvMetadata" @change="setPreviousIsValid(false)">
```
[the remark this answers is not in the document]
brianignacio5: Did you click on the save settings button below? It should pick the value from the input element

src/utils.ts
brianignacio5: It should have a default value on vscode but added default here.

src/views/onboarding/UsePrevious.vue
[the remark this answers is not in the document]
brianignacio5: Oh I get it. The thing is, if you modify user level settings and you have workspace settings, when it reads the value, it goes from workspace to user settings so you would always see the value from workspace settings.

src/examples/ExamplesPanel.ts
```diff
@@ -82,8 +83,8 @@ export class ExamplesPlanel {
                     selectedFolder[0].fsPath,
                     message.name
                   );
-                  ensureDir(resultFolder).then(() => {
-                    utils.copyFromSrcProject(
+                  ensureDir(resultFolder).then(async () => {
+                    await utils.copyFromSrcProject(
```
SoumeshBanerjee, line 87: Will this block the UI / vscode extension/ extension host if the example is significantly big .... say 200mb+ ?

src/newProject/NewProjectPanel.ts
```diff
@@ -0,0 +1,335 @@
...
+    if (NewProjectPanel.currentPanel) {
+      NewProjectPanel.currentPanel.panel.reveal(column);
+    } else {
+      NewProjectPanel.currentPanel = new NewProjectPanel(
+        extensionPath,
+        newProjectArgs,
+        column
+      );
+    }
+  }
```
SoumeshBanerjee, line 55: For me the `Create new Project` is not rendering any UI

Screen Shot 2020-04-27 at 1 12 34 PM

brianignacio5: Any errors in Developer Tools/ Logs/ ESP-IDF channel?

src/examples/ExamplesPanel.ts
brianignacio5: This should not block the UI since there is no await before ensure. Should remove the await from the utils.copyFromSrcProject function. About the waiting time, if the user has built within IDF_PATH/examples then probably. We could ignore build folder in this function to mitigate long time copies.

src/newProject/NewProjectPanel.ts
SoumeshBanerjee: None actually ... which is strange...  I will debug it and let you know soon.

SoumeshBanerjee: I found the root cause of this, is that when `IDF_PATH` is set incorrectly it just crash silently and not render

Possible point of failure:
`public async loadInitialMetadata(metadata: IMetadataFile) {...}`

SoumeshBanerjee: Also crashing here 
```js
const pyVenvList = metadata.venv.filter(...
```

src/newProject/NewProjectPanel.ts
```diff
@@ -0,0 +1,356 @@
...
+            newProjectArgs.metadata.idf &&
+            newProjectArgs.metadata.idf.length > 0 &&
+            newProjectArgs.metadata.venv &&
+            newProjectArgs.metadata.venv.length > 0 &&
+            newProjectArgs.metadata.tools &&
+            newProjectArgs.metadata.tools.length > 0
+          ) {
+            this.loadInitialMetadata(newProjectArgs.metadata);
+          } else {
+            Logger.infoNotify("No values available in metadata.json");
```
pwmb, line 174: For existing user (who are using the extension -- onboarding is done previously) its unable to find meradata file and throwing the error with blank webview

brianignacio5: Hmm good point, should we also include the current values of `idf.customExtraPaths`, `idf.pythonBinPath` and `idf.espIdfPath` to build an initial metadata.json if not available ?

pwmb: probably or ask the user to generate one himself or auto-generate by showing a popup like `Metadat file don't exists and do you like to auto generate or custom generate it`

pwmb:
While running the `New Project Wizard` facing some issues:

```bash
Bin path is not valid for xtensa-esp32-elf
Bin path is not valid for xtensa-esp32s2-elf
Bin path is not valid for openocd-esp32
/bin/sh: : command not found
Command failed: "" "/Users/soumesh/esp/esp-idf/tools/check_python_dependencies.py" -r "/Users/soumesh/esp/esp-idf/requirements.txt"
/bin/sh: : command not found
```

Screen Shot 2020-06-23 at 5 51 47 PM

brianignacio5:
Well, the idea is to be able to handle multiple IDF Versions with corresponding IDF Tools, so having metadata in single project beats the purpose. You can choose where to save settings from previous tools so you can move between versions in current project directory. Custom projects settings are still using vscode settings.json model.

I agree that should not save metadata.json in extension folder. Maybe can use 
* Mac/Linux : $HOME/.vscode/extensions/esp_idf_extension_metadata.json 
* Windows : $USERPROFILE/.vscode/extensions/esp_idf_extension_metadata.json

Will take a look at New Project Wizard issues.

brianignacio5:
> While running the `New Project Wizard` facing some issues:
> 
> ```shell
> Bin path is not valid for xtensa-esp32-elf
> Bin path is not valid for xtensa-esp32s2-elf
> Bin path is not valid for openocd-esp32
> /bin/sh: : command not found
> Command failed: "" "/Users/soumesh/esp/esp-idf/tools/check_python_dependencies.py" -r "/Users/soumesh/esp/esp-idf/requirements.txt"
> /bin/sh: : command not found
> ```
> 
> Screen Shot 2020-06-23 at 5 51 47 PM
> 
> Checked in the OS' terminal (paths are actually correct)

Can you give some steps to reproduce? It's working for me. The command failed seems to not getting the pythonBinPath from the system. Maybe some issue finding a default shell ?

pwmb:
> Can you give some steps to reproduce? It's working for me. The command failed seems to not getting the pythonBinPath from the system. Maybe some issue finding a default shell ?

Strange because I had to run the onboarding again to unlock from the error, all other branched were running fine (build, flash, menuconfig etc) but this branch was not launching anything (_but since I ran the onboarding again, everything is fine_) -- Maybe you can try this case on a fresh folder git clone, run onboaring on master then switch branch to this one and check if everything is fine (if so then could be something wrong with my system)

_Coming to another point_

In `Enter ESP-IDF Component directory` can select multiple folders and then when click `create project` it ask to select folder again, and only create the project there, so whats the purpose of this step

Screen Shot 2020-06-24 at 3 49 39 PM


Also after creating project its not opening the created project

brianignacio5:
This step is to add the components folder which are going to be copied to the created folder. Then it ask for project name and where to save the project.

src/newProject/NewProjectPanel.ts
```diff
@@ -0,0 +1,414 @@
...
+    ) {
+      allIsValid = false;
+    }
+    this.panel.webview.postMessage({
+      command: "idf_tools_check_done",
+      allIsValid,
+    });
+  }
+
+  public async createProject(
```
pwmb, line 330: call this inside a `vscode.window.withProgress(...)` to show a progress bar to the user.

PR #59, opened 2020-03-17
Add Prettier to manage linting and formatting  (VSC-264)

SoumeshBanerjee:
Prettier will be run as `husky` `pre-commit` hook and will format changed files in that commit

Prettier can be installed in `vscode` and can be use with `format-on-save` to help duroing the development of code

Prettier has been added to `Github Action` for `CI`, it checks all the files (`$ yarn prettier --check "**/*.*"`) except `.git/` and all the others mentioned in `.prettierignore` file


.prettierrc.yaml
```diff
@@ -0,0 +1,3 @@
+tabWidth: 2
+semi: true
```
brianignacio5: So I think this is the part that divides developers, those who like 4, 2 whatever spaces tab and those { at the end of the line and its own line hahaha

brianignacio5: LGTM.

PR #62, opened 2020-03-20
Support for ESP-IDF Integrated terminal (VSC-266)

brianignacio5:
Add `ESP-IDF: Open ESP-IDF Terminal` command to launch a new terminal with preloaded IDF_PATH and PATH with ESP-IDF Tools and Python environment executable as well as other tools such as in `IDF_PATH/export.sh`


src/extension.ts
```diff
@@ -847,6 +849,28 @@ function createMonitor(): any {
...
+        espIdfTerminal.show();
+        espIdfTerminal.sendText(`${envSetCmd} IDF_PATH=${modifiedEnv.IDF_PATH}`);
+        let IDF_ADD_PATHS_EXTRAS =
+            `${path.join(modifiedEnv.IDF_PATH, "components", "esptool_py", "esptool")}`;
+        IDF_ADD_PATHS_EXTRAS =
+            `${IDF_ADD_PATHS_EXTRAS}${path.delimiter}${path.join(modifiedEnv.IDF_PATH,
+                                                                 "components", "espcoredump")}`;
+        IDF_ADD_PATHS_EXTRAS =
+            `${IDF_ADD_PATHS_EXTRAS}${path.delimiter}${path.join(modifiedEnv.IDF_PATH,
+                                                                 "components", "partition_table")}`;
```
igrr: Note that these paths may change between IDF versions. This is probably okay for now, but we need to find some way to communicate such path to the IDEs. Have logged this as IDF-1501.

src/extension.ts
```diff
@@ -847,6 +849,28 @@ function createMonitor(): any {
...
+        espIdfTerminal.sendText(`${envSetCmd} PATH=${IDF_ADD_PATHS_EXTRAS}${path.delimiter}${modifiedEnv.PATH}`);
```
igrr: PATH environment variable (at least on Windows) is very likely to contain spaces. You need to add quotes, differently for bash and for Windows CMD:
```
export PATH="${IDF_ADD_PATHS_EXTRAS}${path.delimiter}${modifiedEnv.PATH}"
set "PATH=${IDF_ADD_PATHS_EXTRAS}${path.delimiter}${modifiedEnv.PATH}"
```
```

File: src/extension.ts [hunk 3]
```diff
@@ -847,6 +849,28 @@ function createMonitor(): any {
     });
 }
 
+function createIdfTerminal() {
+    PreCheck.perform([webIdeCheck, openFolderCheck], () => {
+        const envSetCmd = process.platform === "win32" ? "set" : "export";
+        const modifiedEnv = utils.appendIdfAndToolsToPath();
```
igrr: also, question, why are you adding esptool, espcoredump, partition_table paths here, and not in appendIdfAndToolsToPath function? Looks like that might be easier to keep all extra paths in one place, next to `path.join(modifiedEnv.IDF_PATH, "tools")`.

igrr:
LGTM otherwise, looks like a very useful feature!

src/extension.ts
```diff
@@ -856,16 +856,7 @@ function createIdfTerminal() {
...
-        let IDF_ADD_PATHS_EXTRAS =
-            `${path.join(modifiedEnv.IDF_PATH, "components", "esptool_py", "esptool")}`;
-        IDF_ADD_PATHS_EXTRAS =
-            `${IDF_ADD_PATHS_EXTRAS}${path.delimiter}${path.join(modifiedEnv.IDF_PATH,
-                                                                 "components", "espcoredump")}`;
-        IDF_ADD_PATHS_EXTRAS =
-            `${IDF_ADD_PATHS_EXTRAS}${path.delimiter}${path.join(modifiedEnv.IDF_PATH,
-                                                                 "components", "partition_table")}`;
-        espIdfTerminal.sendText(`${envSetCmd} PATH=${IDF_ADD_PATHS_EXTRAS}${path.delimiter}${modifiedEnv.PATH}`);
+        espIdfTerminal.sendText(`${envSetCmd} IDF_PATH="${modifiedEnv.IDF_PATH}"`);
```
igrr: Keep in mind that for cmd.exe the quoting syntax is
```
set "PATH=..."
```
unlike export syntax in bash:
```
export PATH="..."
```

src/utils.ts
```diff
@@ -475,13 +475,22 @@ export function appendIdfAndToolsToPath() {
     }
 
     const idfPathDir = idfConf.readParameter("idf.espIdfPath");
-    modifiedEnv.IDF_PATH = idfPathDir || process.env.IDF_PATH;
+    modifiedEnv.IDF_PATH = `"${idfPathDir}"` || `"${process.env.IDF_PATH}"`;
```
igrr: The only case you need to quote IDF_PATH is if it appears as an argument of the set/export command (`export IDF_PATH="..."`). If you are directly setting its value in the environment variable mapping, then quotes re not needed.

igrr: By the way, for now in the CMake build system, we don't support spaces in IDF_PATH. So you might not need quoting here at all.

src/extension.ts
brianignacio5: Oh thanks, sorry about that.

src/utils.ts
brianignacio5: Understood. For some reason we still need to do set/export in terminal since vscode API doesn't override existing IDF_PATH. 

Quoting on set/export command to keep it compatible.

PR #73, opened 2020-03-28
Enhance project structure (VSC-277)

SoumeshBanerjee:
Enhance Project Structure by:
- Removing some of the unused files
- Moving the logo to the media folder
- Ignore `gulp` script in `vsix`


brianignacio5:
Should we have prettier for Python files or have bandit/flake in the subdirectory ?

SoumeshBanerjee:
> Should we have prettier for Python files or have bandit/flake in the subdirectory ?

Well `prettier` don't have python support and in their github they suggest to use [black](https://github.com/psf/black). I will try it and if its good we can add this to format our python code 

but as of not these bandid/flake8 files which i removed didn't do anything since they were not run in CI or in the npm scripts as well and both the script also have added `esp_debug_adapter` to ignore list

kumekay:
> Well `prettier` don't have python support and in their github they suggest to use [black]

We use [yapf](https://github.com/google/yapf)  to format code in all other projects (and here too https://github.com/espressif/vscode-esp-idf-extension/blob/master/.style.yapf ), so it's better to stick with it in this project too.

SoumeshBanerjee:
> We use yapf to format code in all other projects

Correct, haven't modified that file, just some file which were un-used in the repo

PR #86, opened 2020-04-21
update tasks with relative paths (VSC-308)

### LINKED ISSUE #72: Build task problems in the Problems tab have unresolved filepaths (VSC-276)
**Describe the bug**
After running the build task, the problems tab is populated with entries parsed from the output. 
However, clicking one of these results in a VS Code error: 

> Unable to open 'main.cpp': Unable to read file '/../main/main.cpp' (Error: Unable to resolve non-existing file '/../main/main.cpp').

Where the actual filepath is '/home/boarchuz/Projects/MyProject/main/main.cpp'

**To Reproduce**
In a project with any error, "Tasks - Run task" -> "Build - Build project". Click the resulting entry in the Problems tab.

**Solution**
The [default problemMatcher config](https://github.com/espressif/vscode-esp-idf-extension/blob/d1f5fd16b3e574a5e95726b93e7660aae915e760/templates/.vscode/tasks.json#L18) has "fileLocation": "absolute" while the output provides relative paths.

Example build output error:
> ../main/main.cpp: In function 'void app_main()':
> ../main/main.cpp:82:7: error: 'some_undeclared_variable' was not declared in this scope
>      ++some_undeclared_variable;
>        ^~~~~~~~~~~~~~~~~~~~~~~~

This can be solved by changing "fileLocation" to the following:
`"fileLocation": ["relative", "${workspaceFolder}/build"],`

The only thing I'm not 100% sure about is if that path is ideal. ie. Which path is the build output's "../main/main.cpp" actually relative to? Using "fileLocation": ["relative", "${workspaceFolder}"] would also work with an altered regexp to handle the "../".

There are many other tasks that seem to use the same problemMatcher config that may be affected too.

**Environment**
Linux Mint 19.3, VS Code 1.43.1, IDF Extension 0.1.4, IDF 'master' (v4.2-dev-701-g0ae960f2f)

brianignacio5:
Fix #72 

PTAL @boarchuz @SoumeshBanerjee 

Tested a bit with Windows and MacOS, seems to be ok for build and flash tasks.

### DIFF
--- a/templates/.vscode/tasks.json
+++ b/templates/.vscode/tasks.json
@@ -13,18 +13,32 @@
                     "PATH": "${config:idf.customExtraPaths}"
                 }
             },
-            "problemMatcher": {
-                "owner": "cpp",
-                "fileLocation": ["relative", "${workspaceFolder}/build"],
-                "pattern": {
-                    "regexp": "^(.*):(\\d+):(\\d+):\\s+(warning|error):\\s+(.*)$",
-                    "file": 1,
-                    "line": 2,
-                    "column": 3,
-                    "severity": 4,
-                    "message": 5
+            "problemMatcher": [
+                {
+                    "owner": "cpp",
+                    "fileLocation": ["relative", "${workspaceFolder}"],
+                    "pattern": {
+                        "regexp": "^\\.\\.(.*):(\\d+):(\\d+):\\s+(warning|error):\\s+(.*)$",
+                        "file": 1,
+                        "line": 2,
+                        "column": 3,
+                        "severity": 4,
+                        "message": 5
+                    }
+                },
+                {
+                    "owner": "cpp",
+                    "fileLocation": "absolute",
+                    "pattern": {
+                        "regexp": "^[^\\.](.*):(\\d+):(\\d+):\\s+(warning|error):\\s+(.*)$",
+                        "file": 1,
+                        "line": 2,
+                        "column": 3,
+                        "severity": 4,
+                        "message": 5
+                    }
                 }
-            },
+            ],
             "group": {
                 "kind": "build",
                 "isDefault": true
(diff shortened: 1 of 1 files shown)


mitchjs:
question, when i do run tasks, build... this gets used... and works
but what about the icons on the status bar... when i click on the "BUILD" icon
it builds, but the problem matcher doesnt

brianignacio5:
Hi @mitchjs Yes the thing is we have a background process that runs CMake to build your project but so far we haven't been able to connect the output of that build command to the problem matcher without using a language server for Visual Studio Code (since it will collide with C/C++ extension).

We are looking into ways to achieve these, if anyone have some suggestions, feel free to open an issue, write on Spectrum or contribute to the project also.

boarchuz:
Hi @brianignacio5,

This is now working for errors within the project folder, however fails for problems in external files (Linux). 
Here is the relevant build task output from a project with some external component that contains a problem:

> /home/boarchuz/Espressif/Libraries/hulp/hulp.cpp: In function 'esp_err_t hulp_ulp_load(const ulp_insn_t*, size_t, uint32_t, uint32_t)':
> /home/boarchuz/Espressif/Libraries/hulp/hulp.cpp:374:5: error: 'asdfghjkl' was not declared in this scope
>      asdfghjkl;
>      ^~~~~~~~~

It seems to be caught by the _relative_ problemMatcher because it matches that regexp. Clicking on the entry in Problems, I get this error as a result:

> Unable to open 'hulp.cpp': Unable to read file '/home/boarchuz/Espressif/Projects/test_project/build/home/matt/Espressif/Libraries/hulp/hulp.cpp' (Error: Unable to resolve non-existing file '/home/boarchuz/Espressif/Projects/test_project/build/home/matt/Espressif/Libraries/hulp/hulp.cpp').

brianignacio5:
Roger that, seems we still need the absolute file path regex

PR #90, opened 2020-04-27
Code coverage (VSC-312)

brianignacio5:
Add code coverage with `gcovr` using `xtensa-eslp-elf-gcov` from Xtensa Toolchain.

Code Editor Coverage

HTML Reports

There is text editor line highlight for covered/uncovered lines and project gcovr html report within vscode.

Coverage files are generated on those components which has the coverage compile option enabled in CMakeLists.txt. 

This could become messy if we want to include some components source coverage as well.

Steps to test:

1) Run `ESP-IDF: Configure extension` to install gcovr into Python virtual env. `gcovr` is added in Debug Adapter's requirements.txt
2) Create a new project using coverage. For ease, just create a project using the `ESP-IDF gcov example`. [Set up ESP-IDF for App-trace Coverage](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/app_trace.html#setting-up-a-project-for-gcov)
3) Build the project and make sure there is no errors.
4) Connect to openocd server and execute `esp32 gcov dump` to generate data coverage files.
4) Run `ESP-IDF: Add Editor coverage` with source files opened in vscode editor.
5) Run `ESP-IDF: Remove editor coverage` to remove coverage highlight.
6) Run `ESP-IDF: Get HTML Coverage Report for project` to show html report in webview.


src/coverage/coverageService.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+export async function buildHtml(dirPath: string) {
+  const result = await _runCmd(
+    "gcovr",
+    [
+      "--filter",
+      ".",
+      "--filter",
+      "/Users/brian/esp-idf/components/",
```
pwmb, line 59: abs path!!

src/coverage/coverageService.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+export async function buildJson(dirPath: string) {
+  const result = await _runCmd(
+    "gcovr",
+    [
+      "--filter",
+      ".",
+      "--filter",
+      "/Users/brian/esp-idf/components/",
```
pwmb, line 42: same here

src/coverage/coverageService.ts
```diff
@@ -0,0 +1,163 @@
...
+  partialLines: vscode.Range[];
+  uncoveredLines: vscode.Range[];
+  editor: vscode.TextEditor;
+  countPerLines: countRange[];
+  allLines: vscode.Range[];
+}
+
+export async function buildJson(dirPath: string) {
+  const result = await _runCmd(
+    "gcovr",
```
pwmb, line 37: Also directly calling of `gcovr` is not launching even though `pip list` have 
`gcovr           4.2` version !!
(_I am using python 2.7 btw_)

src/coverage/coverageService.ts
```diff
@@ -0,0 +1,163 @@
...
+  return execChildProcess(
+    `${cmd} ${args.join(" ")}`,
+    dirPath,
+    OutputChannel.init(),
+    { env: modifiedEnv, cwd: dirPath }
+  ).catch((e) => {
+    const msg = e.message ? e.message : e;
+    Logger.error("Error on gcov cmd.\n" + msg, e);
+    OutputChannel.appendLine("Error building gcov cmd.\n" + msg);
+    return "";
```
pwmb, line 80: maybe can use `spawn`? 
Although I am not sure since how long the command responses are ? if they are huge, `exec ` will wait for all of the response and return a buffer instead of stream!!

src/coverage/coverageService.ts
brianignacio5: Should add to  python virtual env in `idf.customExtraPaths` in https://github.com/espressif/vscode-esp-idf-extension/pull/104

src/coverage/coverageService.ts
brianignacio5: I had tried with just a couple of files so exec worked pretty well but yeah if the user chooses many sources files for coverage it could truncate the output. Will fix now.

src/coverage/coverageService.ts
```diff
@@ -0,0 +1,163 @@
...
+  });
+}
+
+export async function generateCoverageForEditors(
+  editors: vscode.TextEditor[],
+  dirPath: string
+) {
+  const gcovObj = await buildJson(dirPath);
```
pwmb, line 92: __Some nitpicking:__
For the scenario where, say you have total 5 source file and you have currently 2 of them opened.
Then you run the `Add Editor Coverage` command, it will render coverage report on the editor for 2 of those already opened source files.
But then if you open 3rd source file or close the already opened one and reopen that.
__Rendering of coverage line on the editor will be gone__ to solve this we have to re run the command, which will call `gcovr` again and again for re computing the same coverage report again.

__Suggestion:__
Maybe we can compute the coverage once for a given fileset and cache (and if that changes then only re compute like how idf build works), and `Add Editor Coverage ` will set a global flag and whenever a new editor tab opens it should check the flag and if the coverage for the source exists then paint the coverage data onto the editor!

src/coverage/coverageService.ts
```diff
@@ -0,0 +1,163 @@
...
+
+export async function previewReport(dirPath: string) {
+  try {
+    const reportHtml = await buildHtml(dirPath);
+    const previewPanel = vscode.window.createWebviewPanel(
+      "gcoveragePreview",
+      "Coverage report",
+      vscode.ViewColumn.One
+    );
+    previewPanel.webview.html = reportHtml;
```
pwmb, line 161: Re running the command will launch multiple webviews (is this desired behaviours?)

src/coverage/coverageService.ts
pwmb: 👍

src/coverage/coverageService.ts
pwmb: Also navigating between already opened editors is also loosing the coverage data highlights

src/coverage/renderer.ts
```diff
@@ -0,0 +1,186 @@
...
+  return coverageOptions;
+}
+
+export class CoverageRenderer {
+  private coveredDecoratorType: vscode.TextEditorDecorationType;
+  private partialDecoratorType: vscode.TextEditorDecorationType;
+  private notCoveredDecoratorType: vscode.TextEditorDecorationType;
+  private countDecoratorTypes: vscode.TextEditorDecorationType[];
+  private marginDecoratorType: vscode.TextEditorDecorationType;
+  private workspaceFolder: vscode.Uri;
```
pwmb, line 63: Color scheme is bit confusing, maybe add some tooltip if possible near the left hand corner text showing line numbers

Also by default coverage color we should add bright light green for covered lines since number is already denoting the hit count!

Screen Shot 2020-05-19 at 3 22 35 PM

pwmb:
@brianignacio5 Thanks! Looks great  

Left a few nit picking and suggestions (code wise)

Also maybe add a MD file in docs for details instructions and also add a section in readme plus the commands and config tables too

src/coverage/coverageService.ts
brianignacio5: Yes, this a standard vscode behavior, decorations disappear when editors are hidden.

Will implement a cache system that re-render on `vscode.window. onDidChangeVisibleTextEditors()` event.

src/coverage/renderer.ts
brianignacio5: I made an initial implementation that had partial coverage for line count=1 and full coverage for count > 1 and uncovered for count = 0. We should probably do as gcovr does and implement covered for >90%*MaxLineCount, 90%*MaxLineCount>partial>75%*MaxLineCount and 75%*MaxLineCount<low covered.

brianignacio5:
The documentation is in `docs/COVERAGE.md` and new configuration settings and commands are added to `README.md`. Is there something there to add?

src/coverage/coverageService.ts
brianignacio5: Fixed to singleton behavior.

src/coverage/renderer.ts
pwmb: This could be cumbersome, maybe we can just show covered and un-covered with green and red simply and show the hit count as showing now on left hand side

src/coverage/coverageService.ts
```diff
@@ -0,0 +1,195 @@
...
+
+export async function generateCoverageForEditors(
+  editors: vscode.TextEditor[],
+  dirPath: string
+) {
+  const gcovObj = await buildJson(dirPath);
+  const coveredEditors: textEditorWithCoverage[] = [];
+  for (let editor of editors) {
+    let gcovObjFilePath = "";
+    if (editor.document.fileName.indexOf(dirPath) > -1) {
```
pwmb, line 100: This is causing un-expected behaviour; 

### Steps to reproduce:
- Open the `examples/system/gcovr` project 
- Run command `Show editor coverage `
- Put breakpoint on 101, and open `Readme.md` file

src/coverage/coverageService.ts
```diff
@@ -0,0 +1,195 @@
...
+      const fileParts = editor.document.fileName
+        .replace(dirPath + sep, "")
+        .split(sep);
+      const fileName = fileParts.pop();
+      gcovObjFilePath = join(
+        "build",
+        "esp-idf",
+        fileParts[fileParts.length - 1],
```
pwmb, line 108: Causing a crash (un handled promise rejection)
Since this is an `async` function its advised to have rejection handler or `try/catch` block for runtime exceptions

src/coverage/coverageService.ts
```diff
@@ -0,0 +1,195 @@
...
+      OutputChannel.appendLine("Error building gcov cmd.\n" + msg);
+      return "";
+    });
+}
+
+export async function generateCoverageForEditors(
+  editors: vscode.TextEditor[],
+  dirPath: string
+) {
+  const gcovObj = await buildJson(dirPath);
```
pwmb, line 96: For every file opened we still calling build json and sometimes (getting error) (_attached screenshot_)

```bash
rejected promise not handled within 1 second: SyntaxError: Unexpected end of JSON input
extensionHostProcess.js:957
stack trace: SyntaxError: Unexpected end of JSON input
	at JSON.parse (<anonymous>)
	at /Volumes/Data/source-code/vscode-esp-idf-extension-gh/dist/extension.js:40434:21
	at Generator.next (<anonymous>)
	at fulfilled (/Volumes/Data/source-code/vscode-esp-idf-extension-gh/dist/extension.js:40406:58)
	at runMicrotasks (<anonymous>)
	at processTicksAndRejections (internal/process/task_queues.js:85:5)
```

Screen Shot 2020-05-21 at 1 00 24 PM

src/coverage/coverageService.ts
```diff
@@ -0,0 +1,195 @@
...
+  }
+  return coveredEditors;
+}
+
+let gcovHtmlPanel: vscode.WebviewPanel;
+export async function previewReport(dirPath: string) {
+  try {
+    if (gcovHtmlPanel) {
+      return;
+    }
```
pwmb, line 181: We should focus to already rendered webview here!

src/coverage/coverageService.ts
brianignacio5: moved the buildJson call to the renderer object hoping to reduce the amount of times being called.

PR #92, opened 2020-04-30
Enhance/task provider (VSC-315)

brianignacio5:
Create a serial task manager that execute tasks sequentially with callbacks.

Move Build and Flash child process to use the  visual studio Code TaskProvider API so we can capture errors in the Problems tab.

Add problem matchers definitions in package.json.


SoumeshBanerjee:
overall build is failing but notification is showing build success

brianignacio5:
PTAL again after using promises, disposables and task cancellation.

src/extension.ts
```diff
@@ -932,38 +926,29 @@ const build = () => {
...
-            return Logger.errorNotify(
-              "IDF Path or IDF Tools path is invalid or not accessible",
-              error
-            );
-          }
-          if (error.code === "ENOENT") {
-            return Logger.errorNotify(
-              `Make sure you have the build tools installed and set in $PATH`,
-              error
-            );
```
pwmb, line 964: This error code don't need to handled?

brianignacio5:
> Couple of points:
> 
> * No notification for build status (build failed, or passed)
> * Every time I build I opens the build output which can be quite distracting (is there a way to hide it or not show to the user all the time or at least hide it for user who closes the output terminal panel)?

There is a limitation with the vscode.Task API, there is a task end event but we don't know happens inside the task and if it fails or not. Same as there is no control to stop task focus when task starts.

src/extension.ts
brianignacio5: Errors are not captured anymore. They were captured within the build child process.

PR #94, opened 2020-05-08
rm pipfile locks (VSC-317)

brianignacio5:
We should remove the lock with previous versions of `psutil`


pwmb:
@brianignacio5 will this solve the security notification from GH?

brianignacio5:
It should solve it @pwmb

PR #97, opened 2020-05-13
Rainmaker Support (VSC-321)

pwmb:
Enabling support for user to connect with rainmaker cloud from inside vscode and provide them functionality to view and update the existing nodes (_added from mobile app_)

Screen Shot 2020-05-13 at 8 48 14 PM


src/extension.ts
```diff
@@ -852,6 +865,195 @@ export async function activate(context: vscode.ExtensionContext) {
...
+  });
+}
+
+function validateInputForRainmakerDeviceParam(
+  value: string,
+  type: string
+): string {
+  if (type === "bool" && !value.match(/^(true|false)$/i)) {
+    return "Enter valid boolean value (true or false)";
+  }
```
brianignacio5, line 1039: Wouldn't it be better to use `vscode.window.showQuickPick` with true/false for bool type here?

Beside this, everything else LGTM!

PR #108, opened 2020-05-21
add regex to include config on search string (VSC-333)

brianignacio5:
Allow users to make search of sdkconfig configurations with optional **CONFIG_** prefix like:

`BT_ENABLED`
`CONFIG_BT_ENABLED`

will (should) return same result.

### DIFF
--- a/src/views/menuconfig/Menuconfig.vue
+++ b/src/views/menuconfig/Menuconfig.vue
@@ -66,7 +66,12 @@ export default class Menuconfig extends Vue {
 
   get items() {
     if (this.searchString !== "") {
-      return filterItems(this.storeItems, this.searchString.toLowerCase());
+      let searchStrMatch = /^(?:CONFIG_)?(.+)/.exec(this.searchString);
+      let searchMatch =
+        searchStrMatch && searchStrMatch.length > 1
+          ? searchStrMatch[1].toLowerCase()
+          : this.searchString.toLowerCase();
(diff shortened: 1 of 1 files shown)


src/views/menuconfig/Menuconfig.vue
```diff
@@ -66,7 +66,12 @@ export default class Menuconfig extends Vue {
 
   get items() {
     if (this.searchString !== "") {
-      return filterItems(this.storeItems, this.searchString.toLowerCase());
+      let searchStrMatch = /^(?:CONFIG_)?(.+)/.exec(this.searchString);
+      let searchMatch =
+        searchStrMatch && searchStrMatch.length > 1
+          ? searchStrMatch[1].toLowerCase()
+          : this.searchString.toLowerCase();
+      return filterItems(this.storeItems, searchMatch);
```
pwmb, line 74: What about some config which is like `Something_BT_BLE` or `WIRELESS_WIFI_SSID` where user might want to have fuzzy search for all of these not just `CONFIG_` one... 

Maybe we could use 
```js
this.storeItem.filter((item) => item.match(searchString.toLowerCase()))
```

brianignacio5: It should work since regex first group is optional. Is it not ?

pwmb: This is working fine, my bad!

PR #121, opened 2020-06-10
Bugfix/idf tools not available (VSC-360)

### LINKED ISSUE #114: TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined (VSC-351)
Not able to install tools in vs code on a 32 bit machine.

***Screenshot
image

***Environment

OS Name: [Windows 7 32-bit]
VSCode Version: [e.g. 1.45.0]
ESP-IDF Version: [e.g. v4.0.1]
Python Version: [e.g. 3.8 32-bit]

brianignacio5:
throw error for IDF Tools when a version is not found. Capture errors on `downloadToolsInIdfToolsPath` caller.

Add cancellable gui menuconfig loading.

Should block IDF Tools download to fix #114 fo example.

### DIFF
--- a/src/idfToolsManager.ts
+++ b/src/idfToolsManager.ts
@@ -113,6 +113,11 @@ export class IdfToolsManager {
         ) > -1
       );
     });
+    if (!versions || versions.length === 0) {
+      throw new Error(
+        `${pkg.name} doesn't have a compatible version for ${this.platformInfo.platformToUse}`
+      );
+    }
     const linkInfo =
       versions.length > 0
         ? (versions[0][this.platformInfo.platformToUse] as IFileInfo)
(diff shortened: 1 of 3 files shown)


src/idfToolsManager.ts
```diff
@@ -113,6 +113,11 @@ export class IdfToolsManager {
         ) > -1
       );
     });
+    if (!versions || versions.length === 0) {
+      throw new Error(
+        `${pkg.name} doesn't have a compatible version for ${this.platformInfo.platformToUse}`
+      );
+    }
```
pwmb, line 120: Did we add `try/catch` in the caller of this function `obtainUrlInfoForPlatform(...)`, since earlier it wasn't throwing now it is.

brianignacio5: Hmm not in its direct caller but on the root caller.

This function is used for `downloadPackage`, `installPackages`, `installZipPackage` and `installTarPackage`. A call tree looks like

downloadToolsInIdfToolsPath
---> downloadPackages
------> downloadPackage
---> installPackages
------> installZipPackage (a or)
------> installTarPackage (b)

and downloadToolsInIdfToolsPath is called by Onboarding with a `.catch`

PR #122, opened 2020-06-11
Search ESP-IDF documentation site from VSCode Editor (VSC-361)

brianignacio5:
Add `ESP-IDF: Search in documentation...` for selected text in editor context menu.

This will open a browser to search for selected text in ESP-IDF. There is a esp-idf doc version list mapping with fallback to latest with esp-idf target.


src/extension.ts
```diff
@@ -565,6 +565,47 @@ export async function activate(context: vscode.ExtensionContext) {
...
+      idfVersion = `latest/${idfTarget}`;
+    }
+    const currentEditor = vscode.window.activeTextEditor;
+    if (currentEditor) {
+      let selection = currentEditor.document.getText(currentEditor.selection);
+      if (!selection) {
+        const range = currentEditor.document.getWordRangeAtPosition(
+          currentEditor.selection.active
+        );
+        selection = currentEditor.document.getText(range);
```
pwmb, line 597: one potential problem is that .... user can select any text and it will open the docs 

One thing would be better @igrr if we can have a param passed to the doc server like `https://docs.espressif.com/projects/esp-idf/en/${idfVersion}/search.html?q={query}&from=vscode.esp-idf-extension`  ... so that doc could render a page if no ref is found that from vscode you can only query valid api or stuff ...

In advance case we can parse the AST and find refs and then only query certain tokens but may be it won't solve the problem still

package.json
```diff
@@ -471,6 +479,10 @@
       {
         "command": "espIdf.getCoverageReport",
         "title": "%espIdf.getCoverageReport.title%"
+      },
+      {
+        "command": "espIdf.searchInEspIdfDocs",
```
pwmb, line 484: Maybe add a `keybinding` here!

src/extension.ts
brianignacio5: Any idea about this ?

pwmb: One possible thing to do is write a parser to parse and package document nodes into a json file which can then be use to lookup the token and if found will then show the content menu `espIdf.searchInEspIdfDocs` and then open the link. But this can be done in later iterations IMO.

```js
selection = currentEditor.document.getText(range);
  if (await util.existInDoc(selection)) { ... }
}
```
JSON file could be simply an array
```json
tokens = [ ... ]
```

brianignacio5: The problem is that JSON file would be huge, it is basically crawling and indexing the whole esp-idf documentation...

pwmb: True, could leave it be for now, in future releases we can have some kind of optimize/compressed list of valid keyword and search mechanism

PR #123, opened 2020-06-11
validate workspace folders length (VSC-362)

### LINKED ISSUE #116: Add note to documentation about folderless workspaces? (VSC-352)
I had a situation yesterday where the extension would not run any commands.  No matter what was tried, VSCode would report "command not found."

After some digging, I found the error at line 65 in extension.ts:

`let workspaceRoot: vscode.Uri;`

in this case, `vscode.Uri` is `undefined` when the current workspace has no folders (this happens when a workspace is created).

There are bug reports in other projects of this being a problem so I think it's a VSCode issue rather than a vscode-esp-idf-extension problem, but it might be worth updating the onboarding doc to mention that the current workspace must have at least one folder or the extension will not work.

brianignacio5:
Fix #116 

PTAL @bdpdx

### DIFF
--- a/src/espIdf/tracing/tree/appTraceArchiveTreeDataProvider.ts
+++ b/src/espIdf/tracing/tree/appTraceArchiveTreeDataProvider.ts
@@ -62,9 +62,11 @@ export class AppTraceArchiveTreeDataProvider
 
   public populateArchiveTree() {
     this.appTraceArchives = Array<AppTraceArchiveItems>(0);
-    const workspace = vscode.workspace.workspaceFolders
-      ? vscode.workspace.workspaceFolders[0].uri.fsPath
-      : "";
+    const workspace =
+      vscode.workspace.workspaceFolders &&
+      vscode.workspace.workspaceFolders.length > 0
+        ? vscode.workspace.workspaceFolders[0].uri.fsPath
+        : "";
     const traceFolder = join(workspace, "trace");
     if (existsSync(traceFolder)) {
       const traceLists = readdirSync(traceFolder);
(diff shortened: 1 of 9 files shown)


src/extension.ts
```diff
@@ -44,7 +44,6 @@ import { AppTraceTreeDataProvider } from "./espIdf/tracing/tree/appTraceTreeData
 import { ExamplesPlanel } from "./examples/ExamplesPanel";
 import { FlashManager } from "./flash/flash";
 import { createFlashModel } from "./flash/flashModelBuilder";
-import { IdfTreeDataProvider } from "./idfComponentsDataProvider";
```
pwmb, line 47: Thanks I wanted to remove this for sooo looong 😆

src/espIdf/tracing/tree/appTraceArchiveTreeDataProvider.ts
```diff
@@ -62,9 +62,11 @@ export class AppTraceArchiveTreeDataProvider
...
   public populateArchiveTree() {
     this.appTraceArchives = Array<AppTraceArchiveItems>(0);
-    const workspace = vscode.workspace.workspaceFolders
-      ? vscode.workspace.workspaceFolders[0].uri.fsPath
-      : "";
+    const workspace =
+      vscode.workspace.workspaceFolders &&
+      vscode.workspace.workspaceFolders.length > 0
+        ? vscode.workspace.workspaceFolders[0].uri.fsPath
+        : "";
```
pwmb, line 69: I see we use this kind of ternary statements for workspace / folder getting .... think would be a better idea to extract these into the util function !!

PR #126, opened 2020-06-15
Support WS:// communication protocol (VSC-365)

pwmb:
#### This feature will be base for the core-dump and gdb-stub implementation

- Add `ws://` (websocket) protocol to communicate
- Add `idf_monitor.py` wrapper for launching monitor with `ws://` mode


src/extension.ts
```diff
@@ -116,6 +118,12 @@ const webIdeCheck = [
   cmdNotForWebIdeMsg,
 ] as utils.PreCheckInput;
 
+const idfBuildChannel = vscode.window.createOutputChannel("ESP-IDF Build");
+const idfFlashChannel = vscode.window.createOutputChannel("ESP-IDF Flash");
+const coreDumpMonitor = vscode.window.createOutputChannel(
+  "ESP-IDF Core Dump Monitor"
+);
+
 export async function activate(context: vscode.ExtensionContext) {
```
brianignacio5, line 127: Could probably delete build and flash channels since they're not used anymore.

So when MonitorType = Core-dump, does it mean that GDB-Stub events are ignored ? (and vice-versa). From the [panic handler workflow](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/fatal-errors.html#panic-handler) it seems core dump would be executed first and then gdb stub. How should we handle this sequential processes ?

pwmb: So basically idea is that there will be one WS Server and only one client is allowed to connect to it (at any given point) since we can't run multiple monitor together anyways.

Now once we start the WSServer ...will get `.on("started")` event and then will start the monitor with `--ws` mode.

Whenever monitor detects gdb stub and/or core-dump it will send `.on("core-dump-detected", () => { ... })` event for processing 

Once IDE is done processing for gdb-stub/coredump it will close ws server and can optionally close monitor.

Removed the `MonitorType` because was just using it to name the terminal in vscode but can name it something generic IMO

PR #135, opened 2020-06-23
add idf task definition (VSC-374)

### LINKED ISSUE #131: Strange Output (VSC-370)
windows 10 x64 1909
vs code 1.46.1
idf-ext 0.3.0

so, just got the new version, as it updated itself
and did a build of my existing project
and noticed there is no more build output.. it is now in terminal
as its launching a task.. which is fine
but i did notice this in the build output - tasks

**Error: the task 'espressif.esp-idf-extension: ESP-IDF Compile' neither specifies a command nor a dependsOn property. The task will be ignored. Its definition is:
{
    "type": "shell",
    "id": "shell,cmake -G Ninja ..,",
    "problemMatcher": [
        "idfRelative",
        "idfAbsolute"
    ],
    "label": "espressif.esp-idf-extension: ESP-IDF Compile"
}**

everything builds just fine.. just noticed that message...mean anything ??

thanks for the hard work on this extension!

brianignacio5:
Not really an error but was using predefined shell task definition which required some additional fields and had some output in output tasks.

Instead, defined our esp-idf task definition.

Fix #131

### DIFF
--- a/package.json
+++ b/package.json
@@ -710,6 +710,17 @@
         "language": "kconfig",
         "path": "./snippets/kconfig.code-snippets"
       }
+    ],
+    "taskDefinitions": [
+      {
+        "type": "esp-idf",
+        "properties": {
+          "command": {
+            "type": "string",
+            "description": "Command to be executed"
+          }
+        }
+      }
     ]
   },
   "extensionDependencies": [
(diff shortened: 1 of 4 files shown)


brianignacio5:
Before I didn't define a task definition and just used the default `shell` task definition, but this was generating some output in Tasks Output Channel (missing some fields in task definition like dependsOn). To fix this, I just add our own task definition to remove the error/warning message.

PR #139, opened 2020-07-07
add arguments to debug adapter manager (VSC-379)

brianignacio5:
Move elf file to debug adapter bag of options (`IDebugAdapterConfig `) and added extra arguments for debug adapter manager

* Add setupCommands for launch.json configuration to pass multiple `--setup-cmd cmd1` for 

launch.json
```
setupCommands: [
   {
      text: "cmd1",
      description: "some description"
   },
   {
      text: "cmd2",
      description: "some description"
   }
]
```
added some `--postmortem` as debug adapter argument for post mortem debug initialization.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,12 +45,14 @@ jobs:
           name: esp-idf-extension.vsix
           path: esp-idf-extension.vsix
 
+      - name: xvfb
+        run: /usr/bin/Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
+
       - name: Extension Test
-        uses: GabrielBB/xvfb-action@v1.2
-        with:
-          run: yarn test --VERBOSE | tee testing.results.log
+        run: yarn test --VERBOSE 2>&1 | tee testing.results.log
         env:
           CODE_VERSION: "1.41.1"
+          DISPLAY: ":99.0"
 
       - name: Upload testing.results.log
         uses: actions/upload-artifact@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Setup Python ${{ matrix.python-version }}
         uses: actions/setup-python@v1
         with:
-          python-version: 3.5
+          python-version: 3.6
 
       - name: Install Node Dependencies
         run: yarn
@@ -49,7 +49,7 @@ jobs:
         run: /usr/bin/Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
 
       - name: Extension Test
-        run: yarn test --VERBOSE 2>&1 | tee testing.results.log
+        run: yarn test --VERBOSE 2>&1 >> testing.results.log
         env:
           CODE_VERSION: "1.41.1"
           DISPLAY: ":99.0"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,14 +45,12 @@ jobs:
           name: esp-idf-extension.vsix
           path: esp-idf-extension.vsix
 
-      - name: xvfb
-        run: /usr/bin/Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
-
       - name: Extension Test
-        run: yarn test --VERBOSE 2>&1 | tee testing.results.log
+        uses: GabrielBB/xvfb-action@v1.2
+        with:
+          run: yarn test --VERBOSE 2>&1 >> testing.results.log
         env:
           CODE_VERSION: "1.41.1"
-          DISPLAY: ":99.0"
 
       - name: Upload testing.results.log
         uses: actions/upload-artifact@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,9 @@ jobs:
       - name: Extension Test
         uses: GabrielBB/xvfb-action@v1.2
         with:
-          run: yarn test --VERBOSE
+          run: yarn test --VERBOSE | tee testing.results.log
+          env:
+            CODE_VERSION: "1.41.1"
 
       - name: Upload testing.results.log
         uses: actions/upload-artifact@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,8 +49,8 @@ jobs:
         uses: GabrielBB/xvfb-action@v1.2
         with:
           run: yarn test --VERBOSE | tee testing.results.log
-          env:
-            CODE_VERSION: "1.41.1"
+        env:
+          CODE_VERSION: "1.41.1"
 
       - name: Upload testing.results.log
         uses: actions/upload-artifact@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,9 +48,7 @@ jobs:
       - name: Extension Test
         uses: GabrielBB/xvfb-action@v1.2
         with:
-          run: yarn test --VERBOSE 2>&1 >> testing.results.log
-        env:
-          CODE_VERSION: "1.41.1"
+          run: yarn test
 
       - name: Upload testing.results.log
         uses: actions/upload-artifact@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,14 +45,10 @@ jobs:
           name: esp-idf-extension.vsix
           path: esp-idf-extension.vsix
 
-      - name: xvfb
-        run: /usr/bin/Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
-
       - name: Extension Test
-        run: yarn test --VERBOSE >> testing.results.log
-        env:
-          CODE_VERSION: "1.41.1"
-          DISPLAY: ":99.0"
+        uses: GabrielBB/xvfb-action@v1.2
+        with:
+          run: yarn test --VERBOSE
 
       - name: Upload testing.results.log
         uses: actions/upload-artifact@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
         run: /usr/bin/Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
 
       - name: Extension Test
-        run: yarn test --VERBOSE 2>&1 >> testing.results.log
+        run: yarn test --VERBOSE 2>&1 | tee testing.results.log
         env:
           CODE_VERSION: "1.41.1"
           DISPLAY: ":99.0"
--- a/package.json
+++ b/package.json
@@ -736,24 +736,13 @@
                 "description": "Environments variables override in Visual Studio Code launched ESP-IDF Debug Adapter",
                 "default": null
               },
-              "setupCommands": {
+              "initGdbCommands": {
                 "type": "array",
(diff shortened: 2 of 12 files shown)


an-dr:
I would suggest to name it `setupGdbCommands` or `initGdbCommands` and format it as array of strings

```jsonc
setupGdbCommands: [
    "cmd1", // comment 1
    "cmd2" // comment 2
]
```

src/espIdf/debugAdapter/debugAdapterManager.ts
```diff
@@ -81,34 +90,34 @@ export class DebugAdapterManager extends EventEmitter {
...
+        this.target,
+      ];
+      this.isPostMortemDebugMode
+        ? adapterArgs.push("-om", "without_oocd")
+        : adapterArgs.push("-om", "connect_to_instance");
+      if (this.isPostMortemDebugMode) {
+        adapterArgs.push("--postmortem"); // MODIFY for actual esp_dap_adapter argument
+      }
+      for (const setupCmd of this.setupCommands) {
+        adapterArgs.push("--setup-cmd", setupCmd.text); // MODIFY for actual esp_dap_adapter argument
```
an-dr, line 118: i'm going to use  `--gdbinit `cause we are usially use files with that name for these commands

src/espIdf/debugAdapter/debugAdapterManager.ts
```diff
@@ -81,34 +90,34 @@ export class DebugAdapterManager extends EventEmitter {
...
+        "-p",
+        this.port.toString(),
+        "-dn",
+        this.target,
+      ];
+      this.isPostMortemDebugMode
+        ? adapterArgs.push("-om", "without_oocd")
+        : adapterArgs.push("-om", "connect_to_instance");
+      if (this.isPostMortemDebugMode) {
+        adapterArgs.push("--postmortem"); // MODIFY for actual esp_dap_adapter argument
```
an-dr, line 115: I'll pick up the `--postmortem` argument for the adapter, no need to change it in future

src/extension.ts
```diff
@@ -1248,10 +1264,25 @@ export async function activate(context: vscode.ExtensionContext) {
       .on("core-dump-detected", (resp) => {
         // perform core-dump related stuff here
         //once done call .done() to notify the monitor process
+        const projectElfFile = "my-custom.elf";
+        const debugAdapterConfig = {
+          elfFile: projectElfFile,
+          isPostMortemDebugMode: false,
+        } as IDebugAdapterConfig;
+        debugAdapterManager.configureAdapter(debugAdapterConfig);
+        debugAdapterManager.start();
```
pwmb, line 1273: This won't start the debugging programmatically, but rather we have to call 

```js
await vscode.debug.startDebugging(undefined, {
  name: "Core Dump Debug",
  type: "espidf",
  request: "launch",
});
```

which will launch the debug adapter by calling 

`vscode.debug.registerDebugAdapterDescriptorFactory("espidf", { ` __[Line:661](https://github.com/espressif/vscode-esp-idf-extension/pull/139/files?file-filters%5B%5D=.ts#diff-45327f86d4438556066de133327f4ca2R661)__ and there we have to see if its started by `Core Dump Debug` by testing `name` then can parse elf file and `core-dump` elf file

package.json
```diff
@@ -683,6 +683,16 @@
...
+              },
+              "initGdbCommands": {
+                "type": "array",
+                "description": "%esp_idf.initGdbCommands.description%",
+                "items": {
+                  "type": "string",
+                  "description": "%esp_idf.debuggers.text.description%",
+                  "default": ""
+                },
+                "default": []
```
pwmb, line 695: [Question]

Is this required for the gdb-stub ?

src/extension.ts
```diff
@@ -242,8 +242,15 @@ export async function activate(context: vscode.ExtensionContext) {
...
       }
+      const projectName = await getProjectName(workspaceRoot.fsPath);
+      const projectElfFile = `${path.join(
+        workspaceRoot.fsPath,
+        "build",
+        projectName
+      )}.elf`;
       const debugAdapterConfig = {
         currentWorkspace: workspaceRoot,
+        elfFile: projectElfFile,
```
pwmb, line 253: What if elf file doesn't exists ?

src/espIdf/debugAdapter/debugAdapterManager.ts
```diff
@@ -24,14 +24,19 @@ import * as idfConf from "../../idfConfiguration";
 import { Logger } from "../../logger/logger";
 import { appendIdfAndToolsToPath, isBinInPath, PreCheck } from "../../utils";
 import { getProjectName } from "../../workspaceConfig";
+import { EOL, tmpdir } from "os";
+import { outputFile } from "fs-extra";
 
 export interface IDebugAdapterConfig {
+  coreDumpFile: string;
```
pwmb, line 31: Shouldn't this be optional, else we need to provide the `coreDumpFile` everywhere?

package.json
brianignacio5: As far as I know we need to set `target remote $port` to gdbstub to work. The actual setting should not be required in the launch.json

src/extension.ts
brianignacio5: I think the debug adapter does require the elfFile to work. Isn't that the case @an-dr ?

src/espIdf/debugAdapter/debugAdapterManager.ts
brianignacio5: Yes, might have been a typo.

src/extension.ts
pwmb: This is true, but what will happen if user launch debug without build ...ideally we should trigger build but showing an alert works as well.

an-dr: We cant work without elf, bacause in the file we have connections between sources and binary code. Personally me would be glad to have a notification (There is no elf-file or smth) with a button "Build now" in that situation

igrr: Note that if we rebuild the ELF file, it may no longer match the program running on the target.
So if the ELF file doesn't exist, and we rebuild it, we should also upload the program again, before starting to debug.

One way to check this is to compare the SHA256(elf file) with the value of SHA256 printed by the panic handler, before running the GDB Stub. This would require changes on the IDF Monitor side, though...

brianignacio5: I added a simple elf exists check that shows an option to either build and flash or don't. The debug session will not start automatically after flash (i.e. need to press F5 again after flashing).

brianignacio5:
I removed our call to xvfb and prefer using [GabrielBB/xvfb-action@v1.2](https://github.com/marketplace/actions/gabrielbb-xvfb-action) as shown in [VSCode API Continuous Integration](https://code.visualstudio.com/api/working-with-extensions/continuous-integration#github-actions) since they have cleanup scripts. Also make sure that the testing.log is uploaded as artifact. What do you think @pwmb  ?

pwmb:
> I removed our call to xvfb and prefer using [GabrielBB/xvfb-action@v1.2](https://github.com/marketplace/actions/gabrielbb-xvfb-action) as shown in [VSCode API Continuous Integration](https://code.visualstudio.com/api/working-with-extensions/continuous-integration#github-actions) since they have cleanup scripts. Also make sure that the testing.log is uploaded as artifact. What do you think @pwmb ?

Awesome 👍

.github/workflows/ci.yml
```diff
@@ -43,14 +43,10 @@ jobs:
...
-        run: /usr/bin/Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
-
       - name: Extension Test
-        run: yarn test --VERBOSE >> testing.results.log
-        env:
-          CODE_VERSION: "1.41.1"
-          DISPLAY: ":99.0"
+        uses: GabrielBB/xvfb-action@v1.2
+        with:
+          run: yarn test
```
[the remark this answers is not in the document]
brianignacio5: yes, I just moved the `>> testing.results.log` to the package.json `test` script

requirements.txt
```diff
@@ -0,0 +1,2 @@
+gcovr
+websocket_client
```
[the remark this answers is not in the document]
brianignacio5: Will create a command to install requirements.txt. Is it better to install all requirements.txt (ESP-IDF -> Debug Adapter -> Extension) or allow the user choose which one to install ?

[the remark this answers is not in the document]
pwmb: Not for now but what I was thinking is to create a dependency list webview which will list all the met/unmet dependencies from various `requirements.txt`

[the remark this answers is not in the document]
pwmb: > Is it better to install all requirements.txt (ESP-IDF -> Debug Adapter -> Extension) or allow the user choose which one to install ?

Yes or rather a dumb command which will fetch all the dependencies regardless ...if it exists `pip` will handle it anyways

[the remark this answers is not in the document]
brianignacio5: Will add this command in #159 since I did the corresponding refactoring there.

pwmb (commented):
Unable to build project in this branch ... some problem with installing python dependencies 

```bash
/esp/python_env/idf4.3_py3.7_env/bin/python /esp/esp-idf/tools/check_python_dependencies.py -r /vscode-esp-idf-extension/esp_debug_adapter/requirements.txt
The following Python requirements are not satisfied:
psutil>=5.5.1
requests>=2.21.0
typing>=3.6.6
xmlrunner>=1.7.7
Please check the documentation for the feature you are using, or run "/esp/python_env/idf4.3_py3.7_env/bin/python -m pip install -r /vscode-esp-idf-extension/esp_debug_adapter/requirements.txt"
Diagnostic information:
    IDF_PYTHON_ENV_PATH: (not set)
    Python interpreter used: /esp/python_env/idf4.3_py3.7_env/bin/python
    Warning: python interpreter not running from IDF_PYTHON_ENV_PATH
    PATH: 
[REDACTED]
```

brianignacio5:
Did you try to update the python requirements ? Need to run the setup to install debug adapter new requirements.txt

pwmb:
> Did you try to update the python requirements ? Need to run the setup to install debug adapter new requirements.txt

Yes running the setup only where its failing

Also, those dependencies are already present in `/esp/python_env/idf4.3_py3.7_env/bin/python` python bin's PIP

pwmb:
Merging to secondary branch to apply other patches and then finally merge to master

src/espIdf/debugAdapter/debugAdapterManager.ts
```diff
@@ -201,9 +236,41 @@ export class DebugAdapterManager extends EventEmitter {
       "esp_debug_adapter",
       "debug_adapter"
     );
+    this.initGdbCommands = [];
+    this.elfFile = "";
+    if (this.currentWorkspace) {
+      this.elfFile = `${path.join(
+        this.currentWorkspace.fsPath,
+        "build",
+        "project-name"
```
pwmb, line 245: @brianignacio5 this path is hardcoded, which is causing the error... will add the project name to fix it

brianignacio5: If you look closely, this for debugAdapter initialization. On extension.ts, when you call debug adapter, you should be getting the new elf file path. Isn't that the case ?

pwmb: Yes this is for initialization, but I was thinking that instead of calling `debugAdapterManager.configureAdapter ` we can have the elf file by default, but anyways we have to call `debugAdapterManager.configureAdapter ` so its fine this way

PR #151, opened 2020-08-05
Add eFuse Explorer (read-only) (VSC-412)

pwmb:
eFuse Bits Explorer will enable users to view all the eFuse Bits for ESP32 & ESP32-S2

### UI Screenshot from VSCode

Screen Shot 2020-08-06 at 7 44 55 PM
Screen Shot 2020-08-06 at 7 46 10 PM
Screen Shot 2020-08-06 at 7 44 34 PM


brianignacio5 (changes requested):
I've tested the feature but I couldn't make it work.

Got this error:

```
Error: non zero exit code 1

Connecting.....
Detecting chip type... ESP32
espefuse.py v3.0-dev
Saving efuse values to /var/folders/m3/q1l25s09679cwww66g3_7m7m0000gn/T/espefusejsondump.tmp\nTraceback (most recent call last):
  File "/Users/brian/esp-idf/components/esptool_py/esptool/espefuse.py", line 183, in <module>
    _main()
  File "/Users/brian/esp-idf/components/esptool_py/esptool/espefuse.py", line 176, in _main
    main()
  File "/Users/brian/esp-idf/components/esptool_py/esptool/espefus…in summary
    json.dump(json_efuse,args.file,sort_keys=True,indent=4)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/json/__init__.py", line 180, in dump
    fp.write(chunk)\nTypeError: a bytes-like object is required, not 'str'

   at ChildProcess.<anonymous> (/Users/brian/workspace/github-vscode-esp-idf-extension/dist/extension.js:72432:28)
   at ChildProcess.emit (events.js:203:13)
   at Process.onexit (internal/child_process.js:272:12)
```
 Here is the port  configuration value `"idf.port": "/dev/cu.usbserial-1433401"` and flash works.

pwmb:
> `Failed to get the eFuse Summary from the chip, please make sure you have selected a valid port`

Was the port valid ?

Also try to clone the latest master of `esptool` repo (gitlab) and replace it in your `IDF_PATH`

brianignacio5:
> Try to clone the latest master of `esptool` repo (gitlab) and replace it in your `IDF_PATH`

Oh right ! Tested it using the esptool master. We should probably add ESP-IDF version validation to make sure that this feature is only executed for v4.3.x or higher (when the esptool module is included in ESP-IDF)

pwmb:
> We should probably add ESP-IDF version validation to make sure that this feature is only executed for v4.3.x or higher

Yeah will add this check. Thanks for suggesting

pwmb:
> > We should probably add ESP-IDF version validation to make sure that this feature is only executed for v4.3.x or higher
> 
> Yeah will add this check. Thanks for suggesting

@brianignacio5 I am thinking that we can't check for sure because user can use some version (which is not official tag) and then won't be able to use the feature ...  instead we can just try to see if the command executed correctly and json received match certain schema if so we can allow ... since we don't have api version in json and can't embed because of the format of data we get `[{...}]` (array of objs) and we need to maintain backward compatibility

brianignacio5 (approved):
LGTM

pwmb:
Prereleased in [`IDF v4.2-beta1`](https://github.com/espressif/esp-idf/releases/tag/v4.2-beta1)

PR #157, opened 2020-08-26
update connection-check key (VSC-424)

an-dr:
This MR adds small fix actualizing debug adapter test mode keys with future release of the Adapter. 

Should be merged AFTER we move to a new adapter released on github as a separated repo


brianignacio5:
Closing this PR since it is fixed in #163

PR #163, opened 2020-09-10
Update debug adapter (VSC-429)

an-dr:
- Updated debug adapter with a new version with coredump and postmortem support
- Re-added adapter as a submodule (https://github.com/espressif/esp-debug-adapter)
- Updated arguments in the adapter test


brianignacio5 (approved):
LGTM

pwmb (approved):

pwmb:
@an-dr Please update the CI for python dependency download step.

One question, not now, but in future once we have #139 merged, can we add the debug adapter as pip package instead of sub-module ? Since #139 have a provision to maintain extension's own `requirements.txt` file

pwmb (changes requested):
Please update the CI for python dependency download step.

pwmb:
@an-dr One question, not now, but in future once we have #139 merged, can we add the debug adapter as pip package instead of sub-module ? Since #139 have a provision to maintain extension's own `requirements.txt` file

pwmb (approved):
LGTM

PR #166, opened 2020-09-14
Bugfix for webview background color issue

### LINKED ISSUE #164: [Bug Report]: Background color broken (VSC-432)
After last update of VSCode from 1.48.2. to 1.49.0 I found that extension`s backgroung has ben broken.
Now all text almost invisible (including project configurator)
screenshot

pwmb:
Previously webview was using `background: transparent` but now from vscode `v1.49.x` need to explicitly specify the background color for the body based on the theme selected by the users

Closes #164


brianignacio5 (approved):
LGTM

PR #170, opened 2020-09-16
Partition Table Editor UI (VSC-439)

pwmb:
Add UI for editing and flashing Partition Table

#### Tasks

- [x] Add support for csv & bins files
- [x] Add Save to write to fs (csv)
- [x] Load content from csv on startup
- [x] Update Docs steps and Troubleshoot
- [x] Remove the bin r/w
- [x] Add input validation
- [x] Add `i18n`
- [x] Add `build & flash` button
- [ ] Once approve add final screen shot in docs

#### Screenshots

Screen Shot 2020-10-21 at 12 19 55 PM

PR #195, opened 2020-10-13
Add new project compiler path

### LINKED ISSUE #191: [Bug Report]: IntelliSense not working properly   (VSC-457)
**Describe the bug**
When creating a new project from the template-app, the file .vscode\c_cpp_properties.json doesn't contain a line for the include files that are included with the compiler of choice.  See the following screenshot.

vscodeerror

**To Reproduce**

1.) Press F1 and type ESP-IDF: Create ESP-IDF project
2.) Give the IDE any path in which to create the project
3.) Select template-app as the template to create the project
4.) Compile the project in order to update compile_commands.json
5.) Observe the include line for FreeRTOS.h, it should look like it does in the attached screenshot.

**Expected behavior**
The file .vscode\c_cpp_properties.json should have a line for the compiler's include directory such as:
```
"C:/Users/[user]/.espressif/tools/xtensa-esp32-elf/esp-2020r2-8.2.0/xtensa-esp32-elf/xtensa-esp32-elf/include/",
```
It would be improved if it could automatically choose the correct compiler include directory (ESP32 vs ESP32-S2 vs ...), and if it could use environment variables to make it more portable.  I have not identified any appropriate environment variables that are already in use by the extension.

- OS Name & Version: [e.g. MacOS X 10.14.5]
- VSCode Version: [e.g. 1.35.1]
- ESP-IDF Version: [e.g. v3.2.2]
- Python Version: [e.g. 2.7.10]

brianignacio5:
Add `compilerPath` to settings.json `C_Cpp.default.compilerPath` for `c_cpp_properties.json` default compiler path on Create project from example.

Fix #191

### DIFF
--- a/src/examples/ExamplesPanel.ts
+++ b/src/examples/ExamplesPanel.ts
@@ -92,7 +92,6 @@ export class ExamplesPlanel {
               canSelectMany: false,
             });
             if (!selectedFolder) {
-              vscode.window.showInformationMessage("No folder selected");
               return;
             }
             try {
--- a/src/examples/ExamplesPanel.ts
+++ b/src/examples/ExamplesPanel.ts
@@ -101,7 +101,10 @@ export class ExamplesPlanel {
                 message.name
               );
               await ensureDir(resultFolder);
-              utils.copyFromSrcProject(message.project_path, resultFolder);
+              await utils.copyFromSrcProject(
+                message.project_path,
+                resultFolder
+              );
               const settingsJsonPath = path.join(
                 resultFolder,
                 ".vscode",
--- a/src/examples/ExamplesPanel.ts
+++ b/src/examples/ExamplesPanel.ts
@@ -82,41 +82,45 @@ export class ExamplesPlanel {
 
     this.panel.onDidDispose(() => this.dispose(), null, this.disposables);
 
-    this.panel.webview.onDidReceiveMessage((message) => {
+    this.panel.webview.onDidReceiveMessage(async (message) => {
       switch (message.command) {
         case "openExampleProject":
           if (message.project_path && message.name) {
-            vscode.window
-              .showOpenDialog({
-                canSelectFolders: true,
-                canSelectFiles: false,
-                canSelectMany: false,
-              })
-              .then((selectedFolder: vscode.Uri[]) => {
-                if (selectedFolder && selectedFolder[0].fsPath) {
-                  const resultFolder = path.join(
-                    selectedFolder[0].fsPath,
-                    message.name
-                  );
-                  ensureDir(resultFolder)
-                    .then(() => {
-                      utils.copyFromSrcProject(
-                        message.project_path,
-                        resultFolder
-                      );
-                      const projectPath = vscode.Uri.file(resultFolder);
-                      vscode.commands.executeCommand(
-                        "vscode.openFolder",
-                        projectPath
-                      );
-                    })
-                    .catch((err) => {
-                      Logger.errorNotify("Error copying ESP-IDF example", err);
-                    });
-                } else {
-                  vscode.window.showInformationMessage("No folder selected");
-                }
-              });
+            const selectedFolder = await vscode.window.showOpenDialog({
+              canSelectFolders: true,
+              canSelectFiles: false,
+              canSelectMany: false,
+            });
+            if (!selectedFolder) {
+              vscode.window.showInformationMessage("No folder selected");
+              return;
+            }
+            try {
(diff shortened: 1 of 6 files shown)


src/utils.ts
```diff
@@ -198,6 +198,19 @@ export function readFileSync(filePath) {
...
+export function readJson(jsonPath: string) {
+  return readJSON(jsonPath);
+}
+
+export function writeJson(jsonPath: string, object: any) {
+  return writeJSON(jsonPath, object, {
+    spaces: vscode.workspace.getConfiguration().get("editor.tabSize") || 2,
+  });
+}
+
```
pwmb, line 213: Is this wrapper really required ? Since we are not adding any extra functionality here, plus these functions are also not getting called from multiple places.

pwmb (changes requested):

templates/.vscode/settings.json
```diff
@@ -7,5 +7,6 @@
   "[c]": {
     "editor.quickSuggestions": true
   },
-  "C_Cpp.intelliSenseEngineFallback": "Enabled"
+  "C_Cpp.intelliSenseEngineFallback": "Enabled",
```
pwmb, line 10: Why are we enabling the `C_Cpp.intelliSenseEngineFallback` flag ?

pwmb (changes requested):

src/examples/ExamplesPanel.ts
```diff
@@ -82,41 +82,48 @@ export class ExamplesPlanel {
...
-                  vscode.window.showInformationMessage("No folder selected");
-                }
-              });
+            const selectedFolder = await vscode.window.showOpenDialog({
+              canSelectFolders: true,
+              canSelectFiles: false,
+              canSelectMany: false,
+            });
+            if (!selectedFolder) {
+              vscode.window.showInformationMessage("No folder selected");
```
pwmb, line 95: Also IMO, this is redundant, since if the user don't select anything we should not perform anything and return silently, no need to be verbose.

src/examples/ExamplesPanel.ts
```diff
@@ -82,41 +82,48 @@ export class ExamplesPlanel {
...
+              return;
+            }
+            try {
```
pwmb, line 98: One suggestion here, going forward we have to show very friendly and concise message to the user with an option to open documentation, so that they can see why this error is happening and how they can unblock themselves

Something similar [I tried here](https://github.com/espressif/vscode-esp-idf-extension/pull/170/files#diff-04bba6a35cad1c794cbbe677678a51de13441b7a6ee8592b7b50be1f05c6f626R1401), maybe later we can have a task to extract this into a function like `showErrorMessageWithDocumentation(msg: string, error: Error, documentationLink: URL, meta: any)`

brianignacio5: I understand the approach but what would the documentation to show here ? We would need to list a document with common errors and how to fix them (here for example would be like file permissions or toolchain not found ?)

pwmb: > need to list a document with common errors and how to fix them

Not really sure, but for now what we can do is since we already have the readme in `/docs` we can add a section of _Troubleshooting_ for all the docs there.

And then link any error in a feature to the specific feature docs.

What do you think?

src/examples/ExamplesPanel.ts
```diff
@@ -82,41 +83,62 @@ export class ExamplesPlanel {
...
+              Logger.error(msg, error);
+              const opt = await vscode.window.showErrorMessage(
+                msg,
+                "Show Docs",
+                "Ok"
+              );
+              if (opt === "Show Docs") {
+                vscode.env.openExternal(vscode.Uri.parse(ESP.URL.Docs.README));
+              }
+              return;
```
pwmb, line 140: return not required here.

src/examples/ExamplesPanel.ts
```diff
@@ -82,41 +83,62 @@ export class ExamplesPlanel {
...
+            } catch (error) {
+              const msg = `Error copying ESP-IDF example.`;
+              Logger.error(msg, error);
+              const opt = await vscode.window.showErrorMessage(
+                msg,
+                "Show Docs",
+                "Ok"
+              );
+              if (opt === "Show Docs") {
+                vscode.env.openExternal(vscode.Uri.parse(ESP.URL.Docs.README));
```
pwmb, line 138: LGTM.

But please create a issue for tracking that we replace the URL with proper one at later point of time

pwmb (approved):

PR #159, opened 2020-09-01
Simplify onboarding (VSC-426)

brianignacio5:
This branch is built to simplify user setup window.

* Cancellable downloads
* Express setup: No customization of ESP-IDF Tools and python virtual environment, just allow to choose ESP-IDF folder and install tools on default IDF_TOOLS_PATH `~/.espressif` or `%USERPROFILE%/.espressif`. If the user updates $IDF_TOOLS_PATH
* Custom setup: Allows the user to select the IDF_TOOLS_PATH or manually input each required tool and environment variable.
* Use default settings if existing ESP-IDF, ESP-IDF Tools and virtual environment are found. This will install `Debug Adapter and Extension requirements` in virtual environment found.

PR #200, opened 2020-10-21
Codicons in sdkconfig editor (VSC-467)

brianignacio5:
Replace font awesome icons to prefer vscode codicons in SDKConfig Editor.

Current onboarding still depends on font-awesome and should be replace after #159 is merged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -835,6 +835,7 @@
     "typescript": "^3.9.5",
     "typescript-tslint-plugin": "^0.3.1",
     "vsce": "^1.52.0",
+    "vscode-codicons": "^0.0.12",
     "vscode-nls-dev": "^3.3.1",
     "vscode-test": "^1.0.0",
     "vue-class-component": "^7.1.0",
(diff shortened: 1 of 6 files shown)


pwmb:
Lets move b43cfd5f85560e47268a34071ad245f41c45fbfd from this PR to a new PR, because updating to `codicons` should be done in a separate PR together with all other webviews and pending PR like Onboarding, Partition Table editor can then also use `codicons` as well.

brianignacio5:
Should we remove font awesome in Onboarding ? I ask because #159 will remove Onboarding and already uses vscode-codicons

pwmb:
So we can remove all the `font-awesome` dependencies from our project and search for `fa` or `fas` etc class name and replace with `codicon` classes.

So yes from onboarding, idf_size, menuconfig, heap_tracing etc... 

For onboarding, we can keep the font-awesome and codicons together... but later we can remove from all

pwmb (approved):
LGTM

PR #203, opened 2020-10-23
Feature cmake editor

### LINKED ISSUE #188: [Feature Request]: CMake Editor (VSC-454)
Please refer this: https://github.com/espressif/vscode-esp-idf-extension/issues/179#issuecomment-703208745

Request From #179

brianignacio5:
Allow user to set project CMakelists.txt or ESP-IDF component CMakelists.txt with a group of input options to fill `idf_component_register` on (component CmakeLists.txt) or  ESP-IDF project CMake in project one.

Closes #188

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "onCommand:espIdf.selectConfTarget",
     "onCommand:espIdf.configDevice",
     "onCommand:menuconfig.start",
+    "onCommand:cmakeListsEditor.start",
     "onCommand:espIdf.setDefaultConfig",
     "onCommand:espIdf.selectPort",
     "onCommand:espIdf.setTarget",
@@ -200,6 +201,11 @@
         {
           "when": "resourceFilename == sdkconfig",
           "command": "menuconfig.start",
+          "group": "navigation"
+        },
+        {
+          "when": "resourceFilename == CMakeLists.txt",
+          "command": "cmakeListsEditor.start",
           "group": "navigation"
         }
       ]
@@ -545,6 +551,10 @@
       {
         "command": "examples.start",
         "title": "%espIdf.examples.title%"
+      },
+      {
+        "command": "cmakeListsEditor.start",
+        "title": "%cmakeListsEditor.start.title%"
       },
       {
         "command": "espIdf.setDefaultConfig",
--- a/src/cmake/cmakeEditorPanel.ts
+++ b/src/cmake/cmakeEditorPanel.ts
@@ -77,9 +77,6 @@ export class CmakeListsEditorPanel {
         retainContextWhenHidden: true,
         localResourceRoots: [
           vscode.Uri.file(join(extensionPath, "dist", "views")),
-          vscode.Uri.file(
-            join(extensionPath, "node_modules", "vscode-codicons", "dist")
-          ),
         ],
       }
     );
@@ -94,13 +91,7 @@ export class CmakeListsEditorPanel {
       )
     );
 
-    const fontsPath = panel.webview.asWebviewUri(
-      vscode.Uri.file(
-        join(extensionPath, "dist", "views", "fonts", "codicon.ttf")
-      )
-    );
-
-    panel.webview.html = this.createCmakeListEditorHtml(scriptsPath, fontsPath);
+    panel.webview.html = this.createCmakeListEditorHtml(scriptsPath);
 
     const cmakeListsWatcher = vscode.workspace.createFileSystemWatcher(
       fileUri.fsPath,
@@ -150,10 +141,7 @@ export class CmakeListsEditorPanel {
     CmakeListsEditorPanel.cmakeListsPanels.add(fileUri.fsPath, panel);
   }
 
-  private createCmakeListEditorHtml(
-    scriptPath: vscode.Uri,
-    fontsUri: vscode.Uri
-  ): string {
+  private createCmakeListEditorHtml(scriptPath: vscode.Uri): string {
     return `<!DOCTYPE html>
         <html lang="en">
         <head>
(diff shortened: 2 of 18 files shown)


package.json
```diff
@@ -835,6 +845,7 @@
     "typescript": "^3.9.5",
     "typescript-tslint-plugin": "^0.3.1",
     "vsce": "^1.52.0",
+    "vscode-codicons": "^0.0.12",
```
pwmb, line 848: this should be gone after rebasing, right ?

package.json
```diff
@@ -201,6 +202,11 @@
           "when": "resourceFilename == sdkconfig",
           "command": "menuconfig.start",
           "group": "navigation"
+        },
+        {
+          "when": "resourceFilename == CMakeLists.txt",
```
pwmb, line 207: Please add docs, about how to use this feature.

src/extension.ts
```diff
@@ -1008,6 +1010,33 @@ export async function activate(context: vscode.ExtensionContext) {
     }
   });
 
+  registerIDFCommand("cmakeListsEditor.start", async (fileUri: vscode.Uri) => {
+    if (!fileUri) {
+      return;
+    }
```
pwmb, line 1016: Since this command is not expected to be called directly by user, maybe we can just disable it from `package.json` and just add this handler.

src/cmake/cmakeEditorPanel.ts
```diff
@@ -0,0 +1,168 @@
...
+      "CMakeLists.txt Editor",
+      column,
+      {
+        enableScripts: true,
+        retainContextWhenHidden: true,
+        localResourceRoots: [
+          vscode.Uri.file(join(extensionPath, "dist", "views")),
+          vscode.Uri.file(
+            join(extensionPath, "node_modules", "vscode-codicons", "dist")
+          ),
```
pwmb, line 70: Is this required since from main.ts we already `import "../commons/espCommons.scss"; `

src/cmake/cmakeEditorPanel.ts
```diff
@@ -0,0 +1,168 @@
...
+        join(
+          extensionPath,
+          "node_modules",
+          "vscode-codicons",
+          "dist",
+          "codicon.css"
+        )
+      )
+    );
+
```
pwmb, line 96: Same here

src/cmake/cmakeListsBuilder.ts
```diff
@@ -0,0 +1,115 @@
...
+  const schemaJson = await readJSON(cmakeListsSchemaFile);
+
+  // Component and Project settings are loaded together
+  // Should refactor if there is a way to identify component/project CmakeLists.txt
+  console.log(schemaJson[type]);
+  return schemaJson[type] as CmakeListsElement[];
+  return [].concat(
+    schemaJson.project as CmakeListsElement[],
+    schemaJson.component as CmakeListsElement[]
+  ) as CmakeListsElement[];
```
pwmb, line 49: second return is useless

src/cmake/cmakeListsBuilder.ts
```diff
@@ -0,0 +1,115 @@
...
+  );
+  const doesSchemaExists = await pathExists(cmakeListsSchemaFile);
+  if (!doesSchemaExists) {
+    return;
+  }
+  const schemaJson = await readJSON(cmakeListsSchemaFile);
+
+  // Component and Project settings are loaded together
+  // Should refactor if there is a way to identify component/project CmakeLists.txt
+  console.log(schemaJson[type]);
```
pwmb, line 44: Maybe remove the `console.log` or can use `Logger`

src/cmake/cmakeListsBuilder.ts
```diff
@@ -0,0 +1,115 @@
...
+      const resultStr = cmakeListFileText.match(regex);
+      if (resultStr && resultStr[1].trim().length > 0) {
+        element.value = [resultStr[1].trim()];
+      }
+    }
+  }
+  return cmakeLists;
+}
+
+export async function updateCmakeListFile(
```
pwmb, line 82: One thing I notice, is that this overrides any comment user wrote, maybe we can try to persist comment, else can provide this caution in docs as well as popup for first time

src/cmake/cmakeEditorPanel.ts
```diff
@@ -0,0 +1,168 @@
...
+  CMakeListsType,
+  loadCmakeListBuilder,
+  updateCmakeListFile,
+  updateWithValuesCMakeLists,
+} from "./cmakeListsBuilder";
+
+export class CmakeListsEditorPanel {
+  public static currentPanel: CmakeListsEditorPanel | undefined;
+
+  public static createOrShow(
```
pwmb, line 29: If already an CMake Editor Webview is open and user right-click and open another cmakelist file ... it doesn't re-render the UI.

pwmb (changes requested):
Thanks for this feature.

Looks really nice, left a few minor comments.

src/extension.ts
brianignacio5: The thing is, command is provided as right click option in the explorer/context and it needs a command title and command entry otherwise an error is shown.

pwmb: Ah! I remember now... maybe we can [do something like this](https://github.com/espressif/vscode-esp-idf-extension/blob/f0dc28127f8d17cd51a42d7d3b9440f868c56957/src/extension.ts#L1147) to inform the user that we can't launch the command directly ...

src/extension.ts
```diff
@@ -1036,6 +1038,19 @@ export async function activate(context: vscode.ExtensionContext) {
     }
   });
 
+  registerIDFCommand("cmakeListsEditor.start", async (fileUri: vscode.Uri) => {
+    if (!fileUri) {
+      Logger.errorNotify(
+        "Cannot call this command directly, right click on any CMakeLists.txt file!",
+        new Error("INVALID_COMMAND")
```
pwmb, line 1045: maybe `INVALID_INVOCATION` since its valid command but invocation isn't

pwmb (changes requested):
- In windows Icons are missing ... 
Screen Shot 2020-11-03 at 12 38 07 PM

- Always adding an empty line in the top

- Comments are rearranged... (lets document this behaviour)

pwmb (approved):
LGTM

ma89vik (approved):

PR #205, opened 2020-10-26
mv to vscode codicons

brianignacio5:
I've moved existing Font awesome to use vscode codicons.

There are some icons that are not available in vscode codicons. Please check the replacement and see if they are ok or should we use others.

### DIFF
--- a/src/views/size/index.html
+++ b/src/views/size/index.html
@@ -36,7 +36,7 @@
                     v-on:click="retryClicked"
                   >
                     <span class="icon">
-                      <i class="codicon codicon-redo"></i>
+                      <i class="codicon codicon-refresh"></i>
                     </span>
                   </button>
                 </p>
--- a/src/views/size/index.html
+++ b/src/views/size/index.html
@@ -25,8 +25,8 @@
                     v-on:click="flashClicked"
                     v-bind:disabled="isFlashing"
                   >
-                    <span class="icon is-small">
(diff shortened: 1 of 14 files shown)


src/views/size/index.html
```diff
@@ -35,8 +35,8 @@ <h1 class="title is-size-5-mobile" v-html="title"></h1>
                     title="Retry"
                     v-on:click="retryClicked"
                   >
-                    <span class="icon is-small">
-                      <i class="fas fa-redo"></i>
+                    <span class="icon">
+                      <i class="codicon codicon-redo"></i>
```
pwmb, line 39: Please use `codicon codicon-refresh`

pwmb (approved):
LGTM, thanks

Just left one suggestion

webpack.config.js
```diff
@@ -94,6 +94,9 @@ const webViewConfig = {
           {
             loader: "css-loader",
           },
+          {
+            loader: "resolve-url-loader",
```
pwmb, line 98: Question: Why this one is required?

brianignacio5: Because vscode codicons font is used in codicon.css with a relative path (./codicon.tff) to the css container directory. This loader helps to resolve css url(). I've noticed that when the extension is bundled into vsix, codicons were not rendering properly on Ubuntu.

PR #208, opened 2020-11-05
Fix codicons font in webviews (VSC-474)

brianignacio5:
`vscode-codicons` has a `codicon.css` with a relative reference to `codicon.ttf` which is broken during Webpack bundle. First we tried to solve it with resolve-url-loader which would work on regular web project but no inside a vscode webview, which uses `vscode-webview-resource` URI and their encoding on each URL. To keep consistency, we need to use `asWebviewUri` API to encode this font, which is released in `dist/views/fonts/codicon.ttf`

This PR is to fix font loading issues in webviews.

### DIFF
--- a/src/espIdf/tracing/system-view/panel.ts
+++ b/src/espIdf/tracing/system-view/panel.ts
@@ -86,6 +86,11 @@ export class SystemViewPanel {
     const scriptPath = this.panel.webview.asWebviewUri(
       Uri.file(join(this.extensionPath, "dist", "views", "sysView-bundle.js"))
     );
+    const fontsUri = this.panel.webview.asWebviewUri(
+      Uri.file(
+        join(this.extensionPath, "dist", "views", "fonts", "codicon.ttf")
+      )
+    );
     this.panel.webview.html = `<!DOCTYPE html>
     <html lang="en">
     <head>
@@ -94,6 +99,12 @@ export class SystemViewPanel {
       <title>System View Trace</title>
     </head>
     <body>
+      <style>
+      @font-face {
+          font-family: "codicon";
+          src: url('${fontsUri}') format('truetype');
+      }
+      </style>
       <section id="app"></section>
       <script src="${scriptPath}"></script>
     </body>
--- a/src/espIdf/tracing/system-view/panel.ts
+++ b/src/espIdf/tracing/system-view/panel.ts
@@ -86,11 +86,6 @@ export class SystemViewPanel {
     const scriptPath = this.panel.webview.asWebviewUri(
       Uri.file(join(this.extensionPath, "dist", "views", "sysView-bundle.js"))
     );
-    const fontsUri = this.panel.webview.asWebviewUri(
-      Uri.file(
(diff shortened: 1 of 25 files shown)


pwmb:
So from the webview codicon example, can't we do it like

[`<meta http-equiv="Content-Security-Policy" content="default-src 'none'; font-src ${codiconsFontUri}; style-src ${webview.cspSource} ${codiconsUri};">`](https://github.com/microsoft/vscode-extension-samples/blob/d559763f2b2c940852ac6172e2af012920199b0d/webview-codicons-sample/src/extension.ts#L46)

maybe we can try this one once?

brianignacio5:
I had tried, the content security policy basically just limit the places where html can look for sources like js scripts, fonts, etc. The codicon.css still needs to find codicon.ttf (and since it is relative to the css itself, we could modify the original css itself, but seems it can break if anything changes with vscode-codicons package).

The main issue is the Webpack bundle due to 2 reasons 1) Font in Webpack is output in dist (we use it from there instead of node_modules since we ignore node_modules in vscode vsix) and 2) the codicon.css has a relative path which can be resolve to find original location (with resolve-url-loader) of font (node_modules/vscode-codicons/dist/codicon.ttf), but not the one in dist folder and this path is not understood in webviews if it is not parsed with vscode-webview URI

src/espIdf/tracing/system-view/panel.ts
```diff
@@ -94,6 +99,12 @@ export class SystemViewPanel {
       <title>System View Trace</title>
     </head>
     <body>
+      <style>
+      @font-face {
+          font-family: "codicon";
+          src: url('${fontsUri}') format('truetype');
+      }
+      </style>
```
pwmb, line 107: How about _(if possible)_ we move these instance in the `espCommons.scss` so that we don't have to type this out manually since most of the webviews are importing that scss...?

brianignacio5: I tried to put it in espCommons.scss but the issue is that the fonts URI needs to be encode using `vscode.asWebviewUri()` to work properly inside the webview.

webpack.config.js
```diff
@@ -112,7 +112,7 @@ const webViewConfig = {
             loader: "file-loader",
             options: {
               name: "[name].[ext]",
-              outputPath: "fonts/",
+              outputPath: "fonts",
             },
```
pwmb, line 116: If we remove this dir option then wouldn't this solve this issue `src: url("./codicon.ttf?669d352dfabff8f6eaa466c8ae820e43") format("truetype");` ?

Since it will then output to `./` dir for the webview i.e., `dist/views`

brianignacio5: Still same issue of encoding using `vscode.asWebviewUri()` and the font location will try to be resolved against original location (~node_modules/vscode-codicons/dist/codicon.ttf since it is where codicon.css is) with `resolve-url-loader` or in the same location as entry file : ${viewDir}/main.ts if resolve-url-loader is ignored.

pwmb (approved):
LGTM, thanks for the fix

Issue #37, opened 2020-02-13
extension flashing performance is 4x slower (VSC-242)


syu-source:
**Describe the bug**
Flashing by extension task is almost 4 times slower than terminal command` idf.py -p com flash`. 

**To Reproduce**
Build project, set correct device port.

1. Click Flash button from toolbar at the bottom, watch esp-idf flash output.

```
Wrote 1057824 bytes (685281 compressed) at 0x00010000 in 60.8 seconds (effective 139.1 kbit/s)...
Hash of data verified.
```

2. Open termianl, run  "idf.py.exe -p COMX flash", watch terminal feedback.
```
Wrote 1058064 bytes (685377 compressed) at 0x00010000 in 16.0 seconds (effective 528.1 kbit/s)...
Hash of data verified.
```

**Expected behavior**
Both should flash the device nice and quick.

**Screenshots**
![image](https://user-images.githubusercontent.com/35789043/74416938-d97f4b00-4e80-11ea-906d-f33390970ab5.png)
![image](https://user-images.githubusercontent.com/35789043/74417003-f7e54680-4e80-11ea-910a-272b1745fa25.png)

**Environment (please complete the following information):**
- OS Name & Version: [win10]
- VSCode Version: [0.13]
- ESP-IDF Version: [4.1]
- Python Version: [3.7]
- Hardware: ESP32-DevKitC with ESP32-WROOM-32

**Additional context**

full log from terminal (integrated windows power shell).
```

idf.py.exe -p COM7 flash  
Adding "flash"'s dependency "all" to list of commands with default set of options.
Executing action: all (aliases: build)
Running ninja in directory d:\workspace\vscode\workspace\hello\build
Executing "ninja all"...
[1/3] Performing build step for 'bootloader'
ninja: no work to do.
Executing action: flash
Running ninja in directory d:\workspace\vscode\workspace\hello\build
Executing "ninja flash"...
[1/4] Performing build step for 'bootloader'
ninja: no work to do.
[1/2] cmd.exe /C "cd /D D:\workspace\vscode\esp-idf.v4\components\esptool_py && D:\workspace\vscode\.../workspace/hello/build" -P D:/workspace/vscode/esp-idf.v4/components/esptool_py/run_esptool.cmake" 
[...]

PR #209, opened 2020-11-05
add flash monitor baud rate (VSC-476)

### LINKED ISSUE #207: [Feature Request]: Configurable flash baudrate (VSC-473)
**Background**
It seems to me that when I use the ESP-IDF: Flash command (i.e. <kbd>Ctrl</kbd> + <kbd>E</kbd>  <kbd>F</kbd>) it gets the flash baud rate (the parameter passed to `-b`) from the configured device baudrate. I would like to flash at the fastest possible baud rate (i.e. 2MHz), but would not always want to communicate at that speed (prefer 115200 for comms).

```
> Executing task in folder sensor_platform: /home/cal/.espressif/python_env/idf4.3_py2.7_env/bin/python /home/cal/esp/esp-idf/components/esptool_py/esptool/esptool.py -p /dev/ttyUSB0 -b 2000000 --after hard_reset write_flash --flash_mode dio --flash_freq 40m --flash_size detect 0x8000 partition_table/partition-table.bin 0xe000 ota_data_initial.bin 0x1000 bootloader/bootloader.bin 0x10000 sensor_platform.bin <
```

**Preferred Solution**
A separate configurable parameter in the esp-idf config, picked up by the vscode extension.

**Alternatives Considered**
I tried using the workspace's `tasks.json` and modifying this, but it also seems to me that this is Task: Flash, and not the ESP-IDF: Flash command, so it did not override the built-in method.

I ended up modifying the extension's `/dist/extension.js` on my install and replacing `this.model.baudRate` with a fixed value of a higher baud rate.

**Additional context**
I realize that the configurable parameter would likely be rooted in the esp-idf config, but thought it would be best to start here.
I also note that changing the flash mode, from DIO to QIO in the esp config, did not seem to have an effect on the parameter passed to `--flash_mode` when flashing via the extension command.

brianignacio5:
Add Monitor and Flash baud rate configuration settings for vscode with higher default baud rate.

Fix #207


src/extension.ts
```diff
@@ -1385,7 +1397,7 @@ export async function activate(context: vscode.ExtensionContext) {
   );
   registerIDFCommand("espIdf.launchWSServerAndMonitor", async () => {
     const port = idfConf.readParameter("idf.port") as string;
-    const baudRate = idfConf.readParameter("idf.baudRate") as string;
+    const baudRate = idfConf.readParameter("idf.monitorBaudRate") as string;
```
pwmb, line 1400: rename the variables as well!

pwmb (commented):
Maybe we can use [`CONFIG_ESPTOOLPY_MONITOR_BAUD `](https://github.com/espressif/vscode-esp-idf-extension/issues/37#issuecomment-697306183) as suggested by @igrr 

Also for commands like `idf.py flash monitor`... what shall be the default behaviour ... for user who are setting both the flash baud rate and monitor baud rate they want both to be run in their own baud rate ... right? is this case handled?

brianignacio5:
> Maybe we can use [`CONFIG_ESPTOOLPY_MONITOR_BAUD `](https://github.com/espressif/vscode-esp-idf-extension/issues/37#issuecomment-697306183) as suggested by @igrr
> 
> Also for commands like `idf.py flash monitor`... what shall be the default behaviour ... for user who are setting both the flash baud rate and monitor baud rate they want both to be run in their own baud rate ... right? is this case handled?

For `idf.py flash monitor` I assume you mean in the ESP-IDF Terminal, where the user can specify the baud rate which is probably going to be use for both flash and monitor actions.

pwmb:
> I assume you mean in the ESP-IDF Terminal, where the user can specify the baud rate

No I mean from the command `espIdf.buildFlashMonitor`

brianignacio5:
> > I assume you mean in the ESP-IDF Terminal, where the user can specify the baud rate
> 
> No I mean from the command `espIdf.buildFlashMonitor`

buildFlashMonitor has `idf.flashBaudRate` for the flash step and `idf.monitorBaudRate` for the monitor step.

templates/.vscode/tasks.json
```diff
@@ -142,9 +142,9 @@
         {
             "label": "Monitor: Start the monitor",
             "type": "shell",
-            "command": "${config:idf.pythonBinPath} ${config:idf.espIdfPath}/tools/idf.py -p ${config:idf.port} -b ${config:idf.baudRate} monitor",
+            "command": "${config:idf.pythonBinPath} ${config:idf.espIdfPath}/tools/idf.py -p ${config:idf.port} monitor",
             "windows": {
-                "command": "${config:idf.pythonBinPathWin} ${config:idf.espIdfPathWin}\\tools\\idf.py monitor -p ${config:idf.portWin} -b ${config:idf.baudRate}"
+                "command": "${config:idf.pythonBinPathWin} ${config:idf.espIdfPathWin}\\tools\\idf.py monitor -p ${config:idf.portWin}"
```
pwmb, line 147: maybe we need to pass the monitor baud rate

pwmb (approved):
Thanks @brianignacio5 LGTM...

Left one minor comment

templates/.vscode/tasks.json
brianignacio5: Based on @igrr comment, `idf.py monitor` already picks the value from CONFIG_ESPTOOLPY_MONITOR_BAUD in https://github.com/espressif/esp-idf/blob/8bc19ba893e5544d571a753d82b44a84799b94b1/tools/idf_py_actions/serial_ext.py#L87-L88

pwmb: I see maybe add a doc explaining this also add the priority/preference order of the baud rates for the monitor.

PR #210, opened 2020-11-05
Menu description correction

technosf:
A couple of VS Extension settings misdescribe _MDF_ as _ADF_

This corrects that.


brianignacio5:
Thanks for the fix

PR #220, opened 2020-11-23
Notifications silent mode

### LINKED ISSUE #213: [Feature Request]: Remove the banners (VSC-479)
**Is your feature request related to a problem? Please describe.**
The banner notifications, such as "Build successfully" that appear in the lower right-hand corner are really annoying because they cover up part of the output log and stay for much too long. It interferes with the workflow because many times it is impossible to see what the last lines out output was.

Could you please add a toggle to turn these banners off altogether?

**Describe alternatives you've considered**
The output from the build is sufficient information and banners are not really required. They really just get in the way.

**Additional context**

### LINKED ISSUE #214: [Bug Report]: Build does not switch to terminal output window (VSC-480)
**Pre Bug Report Checklist**
Before reporting any bug please make sure of these points.

- [x] Make sure you have searched for existing bugs and features request before you post an issue.
- [x] This is a bug report for the ESP-IDF Visual Studio Code extension and **not** an ESP-IDF bug report.
- [x] I've read the docs and found no information that could have helped solving the issue.

**Describe the bug**
When starting a build, many times the system will not activate the terminal output window that shows the make log. While the banner displays that the build is in progress, there is no indication as to the actual build output, as it is hidden from view. It only activates it, it appears, if the build fails.

**To Reproduce**
Switch to the window or bash window and build

**Expected behavior**
When the build starts, the corresponding terminal window should be activated and displayed that shows the output from the build script.

**Screenshots**

- OS Name & Version: MacOS X 10.14.6
- VSCode Version: 1.51.1
- ESP-IDF Version: v4.0.2
- Python Version: 3.8

**Logs**

**Additional context**

brianignacio5:
Add `idf.notificationSilentMode ` to disable/enable extension notifications. Build/Flash tasks output will be shown if false.

Fix #214
Fix #213

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ These are the configuration settings that ESP-IDF extension contributes to your 
 | `idf.partialDarkTheme`       | Background color for partially covered lines in dark theme for gcov coverage  |
 | `idf.uncoveredLightTheme`    | Background color for uncovered lines in light theme for gcov coverage         |
 | `idf.uncoveredDarkTheme`     | Background color for uncovered lines in dark theme for gcov coverage          |
-| `idf.notificationSilentMode` | Silent all notifications messages                                             |
+| `idf.notificationSilentMode` | Silent all notifications messages (excluding error notifications)             |
 
 When you use the command **ESP-IDF: Set Espressif device target** it will override `idf.adapterTargetName` with selected chip and `idf.openOcdConfigs` with its default OpenOCD Configuration files. If you want to customize the `idf.openOcdConfigs` alone, you can modify your user settings.json or use **ESP-IDF: Device configuration** and select `Enter OpenOCD Configuration File Paths list` by entering each file separated by comma ",".
 
--- a/README.md
+++ b/README.md
@@ -162,6 +162,7 @@ These are the configuration settings that ESP-IDF extension contributes to your 
 | `idf.partialDarkTheme`     | Background color for partially covered lines in dark theme for gcov coverage  |
 | `idf.uncoveredLightTheme`  | Background color for uncovered lines in light theme for gcov coverage         |
 | `idf.uncoveredDarkTheme`   | Background color for uncovered lines in dark theme for gcov coverage          |
+| `idf.focusTaskOutput`      | Focus on Build/Flash task output on start                                     |
 
 When you use the command **ESP-IDF: Set Espressif device target** it will override `idf.adapterTargetName` with selected chip and `idf.openOcdConfigs` with its default OpenOCD Configuration files. If you want to customize the `idf.openOcdConfigs` alone, you can modify your user settings.json or use **ESP-IDF: Device configuration** and select `Enter OpenOCD Configuration File Paths list` by entering each file separated by comma ",".
 
--- a/README.md
+++ b/README.md
@@ -144,25 +144,26 @@ This extension contributes the following settings that can be later updated in s
 
 These are the configuration settings that ESP-IDF extension contributes to your Visual Studio Code editor settings.
 
-| Setting ID                 | Description                                                                   |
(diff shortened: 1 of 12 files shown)


pwmb (approved):
Thanks for the PR. LGTM!

src/userNotificationManager/userNotificationManager.ts
```diff
@@ -30,12 +31,15 @@ export default class UserNotificationManagerTransport extends winston.Transport
...
+    ) as boolean;
     if (metadata && metadata.user) {
-      if (level === "info") {
+      if (level === "info" && !focusTaskOutput) {
         vscode.window.showInformationMessage(message);
-      } else if (level === "warn") {
+      } else if (level === "warn" && !focusTaskOutput) {
         vscode.window.showWarningMessage(message);
-      } else if (level === "error") {
+      } else if (level === "error" && !focusTaskOutput) {
```
pwmb, line 42: Not on error, let keep error intact, and user should receive error notification always

README.md
```diff
@@ -144,24 +144,26 @@ This extension contributes the following settings that can be later updated in s
...
+| `idf.adapterTargetName`      | ESP-IDF target Chip (Example: esp32)                                          |
+| `idf.openOcdConfigs`         | Configuration files for OpenOCD. Relative to OPENOCD_SCRIPTS folder           |
+| `idf.saveBeforeBuild`        | Save all the edited files before building (default `true`)                    |
+| `idf.coveredLightTheme`      | Background color for covered lines in light theme for gcov coverage           |
+| `idf.coveredDarkTheme`       | Background color for covered lines in dark theme for gcov coverage            |
+| `idf.partialLightTheme`      | Background color for partially covered lines in light theme for gcov coverage |
+| `idf.partialDarkTheme`       | Background color for partially covered lines in dark theme for gcov coverage  |
+| `idf.uncoveredLightTheme`    | Background color for uncovered lines in light theme for gcov coverage         |
+| `idf.uncoveredDarkTheme`     | Background color for uncovered lines in dark theme for gcov coverage          |
+| `idf.focusTaskOutput`        | Focus on Build/Flash task output on start                                     |
```
pwmb, line 165: lets remove `idf.focusTaskOutput ` settings because user, who wants to silent their notification can enable `idf.notificationSilentMode`

README.md
```diff
@@ -144,24 +144,26 @@ This extension contributes the following settings that can be later updated in s
...
+| `idf.notificationSilentMode` | Silent all notifications messages                                             |
```
pwmb, line 166: Update Description add `(excluding error notifications)`

i18n/en/package.i18n.json
```diff
@@ -59,6 +59,8 @@
   "trace.wait4halt.description": "wait4halt will be set for the apptrace",
   "trace.skip_size.description": "skip_size will be set for the apptrace",
   "param.saveBeforeBuildDescription": "Save all the edited files in the workspace before proceeding with the build, although if fail to save files it will build anyways",
+  "param.focusTaskOutput": "Show Build/Flash tasks Output log",
+  "param.notificationSilentMode": "Disable all ESP-IDF extension notifications",
```
pwmb, line 63: Same here (update Description)

pwmb (changes requested):

README.md
brianignacio5: But the purpose of the setting is different, focusTaskOutput is now only to show task output when build/flash happens.

pwmb: true but since `idf.notificationSilentMode ` does the same thing but on a global level, its not required to give settings to silent each and individual notifications (at least for now), we have have simple toggle, i.e., either users will get all notification or none (except for errors because they are important)

pwmb (approved):
LGTM, Thanks

PR #224, opened 2020-12-01
Add WSL serial support

brianignacio5:
Add code to support get serial ports, flash and monitor on WSL.

Tested on Windows 10 19041 with WSL 2 with Microsoft Ubuntu 20.04 distribution for WSL.

### DIFF
--- a/src/espIdf/serial/serialPort.ts
+++ b/src/espIdf/serial/serialPort.ts
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-import { EOL, release } from "os";
+import { release } from "os";
 import * as vscode from "vscode";
 import * as idfConf from "../../idfConfiguration";
 import { LocDictionary } from "../../localizationDictionary";
@@ -130,7 +130,7 @@ export class SerialPort {
       Logger.error(errMSg, error);
     }
     const choices: SerialPortDetails[] = Array<SerialPortDetails>();
-    for (let r of result.toString().trim().split(EOL)) {
+    for (let r of result.toString().trim().split("\r\n")) {
       choices.push(new SerialPortDetails(r));
     }
     return choices;
--- a/src/espIdf/serial/serialPort.ts
+++ b/src/espIdf/serial/serialPort.ts
@@ -16,11 +16,12 @@
  * limitations under the License.
  */
 
+import { EOL, release } from "os";
 import * as vscode from "vscode";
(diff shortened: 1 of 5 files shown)


src/espIdf/serial/serialPort.ts
```diff
@@ -105,4 +116,24 @@ export class SerialPort {
       }
     });
   }
+
+  private async wslList(): Promise<SerialPortDetails[]> {
+    let result = "";
+    try {
+      result = await execChildProcess(
+        `powershell.exe -Command "$(cat wsl_get_serial_list.ps)"`,
```
pwmb, line 124: is this compatible with both old wsl and wsl2 ?

pwmb: also source for this file?

brianignacio5: Should we add the source in the file ? I though the code is small and general enough.

The .NET Framework 2.0 defines the `System.IO.Ports.SerialPort` and should be independent of WSL version.

Will test it soon.

pwmb: I think its better to add source, in case in future we don't remember what code we wrote to generate this bin.

Also it will be much easier for git to track

pwmb (approved):
LGTM, tested on win 10

src/espIdf/serial/serialPort.ts
brianignacio5: I added the source in `wsl_get_serial_list.ps`

zebulon501:
I understand this changes ease the access to some people, but it has a few limitations IMHO:
* Native Windows processes are launched without any warning;
* Python must be installed on Windows, along with pySerial (missing dependency by the way);
* Specifying an RFC2217 serial port doesn't prevent using this mechanism;
* WSL2 will soon offer USBIP (see https://devblogs.microsoft.com/commandline/connecting-usb-devices-to-wsl/), thus bringing native WSL USB ports, like `/dev/ttyUSB0`, which is unusable due to this mechanism;
* Using devcontainer is also broken: this configuration is detected as WSL but has no access to Windows binaries.

This behavior (using powershell.exe and path substitution) shall only be used when the user specifies `idf.port` as `COMx` and running under Linux. Detecting WSL using the kernel information is not working (docker use case), better use the environment variable `WSL_DISTRO_NAME`.

PR #229, opened 2020-12-09
Many small bugfixes

### LINKED ISSUE #223: [Bug Report]: esp-idf-v4.0.2.zip doesn't exist for v4.0 (release branch) (VSC-489)
- OS Name & Version: [Win10Pro x64 10.0.19041.630]
- VSCode Version: [1.51.1]
- ESP-IDF Version: [release/v4.0 (release branch)]
- Python Version: [3.9.0]

If I select (release / v4.0 (release branch)) in the (Configure ESP-IDF) step, then I get an error after downloading the framework archive because vscode-esp-idf-extension is trying to find the archive (esp-idf-v4.0.2.zip) for v4.0.2 release version, but not for v.4.0 release branch, screen below.
I think this is a configuration wizard script error ...
image

### LINKED ISSUE #234: [Feature Request]: Win configuration parameters (VSC-497)
There is no information about *Win configuration parameters like idf.portWin in README. I tried to use configuration parameters without Win and got an error, because it used COM1 by default instead of my COM9. Only pressing F1 and "ESP-IDF: Select port to use" clarified the situation.

Could you please add some kind of information about *Win parameters to README? Maybe I just don't know where to search it :)

### LINKED ISSUE #245: Rainmaker UI Crash
If user is not loggedIn then the UI is crashing.

This change is causing crash if user is not logged in already

_Originally posted by @pwmb in https://github.com/espressif/vscode-esp-idf-extension/pull/233#discussion_r546688785_

brianignacio5:
Only copy .vscode folder files on template if they don't exist on the current project path.

Small fixes such as:
* update idf_versions.txt fallback file.
* update filename for esp-idf version on release (fix #223)
* Fix spaces bug in windows registry python
* Fix Examples command, webview title, rendered html and [security alert fix](https://github.com/espressif/vscode-esp-idf-extension/security/code-scanning/1?query=ref%3Arefs%2Fheads%2Fmaster)
* Rainmaker tree item check undefined tooltip and label (fix #245)
* Add Windows configurations settings in README (fix #234 )


pwmb (approved):

src/extension.ts
```diff
@@ -1006,14 +1014,17 @@ export async function activate(context: vscode.ExtensionContext) {
             const examplesFolder = await vscode.window.showQuickPick(
               [
                 {
+                  description: "ESP-IDF",
```
pwmb, line 1017: I would suggest based on these `espIdfPath`, `espAdfPath` and `espMdfPath` we can just disable these options or not show them at all? What do you think?
Also if we just have only one available then just short circuit this step completely!

pwmb (approved):
Left one suggestion, LGTM

src/extension.ts
brianignacio5: But they are available ! If you had set ESP-MDF or ESP-ADF (with commands or manually set configuration settings) you can see their examples rendered here.

src/rainmaker/view/item.ts
```diff
@@ -48,9 +48,18 @@ export class RMakerItem extends TreeItem {
...
+        typeof this.label === "string"
+          ? this.label
+          : this.label.label
+          ? this.label.label
+          : undefined,
       tooltip:
-        typeof this.tooltip === "string" ? this.tooltip : this.tooltip.value,
+        typeof this.tooltip === "string"
+          ? this.tooltip
+          : this.tooltip.value
```
pwmb, line 60: this will still cause error... because 

```js
if (typeof this.tooltip === "string") {
    tooltip = this.tooltip;
} else if ( this.tooltip.value) {
               //^--- crash here since this.tooltip is undefined
    tooltip = this.tooltip.value;
} else { 
    tooltip = undefined;
}
```

pwmb (changes requested):

src/extension.ts
```diff
@@ -1003,21 +1011,31 @@ export async function activate(context: vscode.ExtensionContext) {
...
+
+            const pickItems = [];
+            if (espIdfPath) {
+              pickItems.push({
+                description: "ESP-IDF",
+                label: `Use current ESP-IDF (${espIdfPath})`,
+                target: espIdfPath,
+              });
+            }
+            if (espAdfPath) {
```
pwmb, line 1023: since `espAdfPath ` is type casted `as string`, the value is `'undefined'` which is string and not `undefined` type and hence `if ` will be `true`

pwmb (changes requested):

pwmb (approved):
LGTM

PR #233, opened 2020-12-14
update yarn lock gulp-cli and vscode api

brianignacio5:
The gulp team has update gulp-cli to fix yargs-parser dependency issue.

By updating the yarn.lock, vscode is update to latest version with minor changes in the TreeView API.

Also fix ini dependency bump

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1132,7 +1132,13 @@ export async function activate(context: vscode.ExtensionContext) {
 
   registerIDFCommand("espIdf.apptrace", () => {
     PreCheck.perform([webIdeCheck, openFolderCheck], async () => {
-      if (appTraceTreeDataProvider.appTraceButton.label.match(/start/gi)) {
+      const appTraceLabel =
+        typeof appTraceTreeDataProvider.appTraceButton.label === "string"
+          ? appTraceTreeDataProvider.appTraceButton.label.match(/start/gi)
+          : appTraceTreeDataProvider.appTraceButton.label.label.match(
+              /start/gi
+            );
+      if (appTraceLabel) {
         await appTraceManager.start();
       } else {
         await appTraceManager.stop();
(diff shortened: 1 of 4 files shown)


src/extension.ts
```diff
@@ -1132,7 +1132,13 @@ export async function activate(context: vscode.ExtensionContext) {
 
   registerIDFCommand("espIdf.apptrace", () => {
     PreCheck.perform([webIdeCheck, openFolderCheck], async () => {
-      if (appTraceTreeDataProvider.appTraceButton.label.match(/start/gi)) {
+      const appTraceLabel =
+        typeof appTraceTreeDataProvider.appTraceButton.label === "string"
+          ? appTraceTreeDataProvider.appTraceButton.label.match(/start/gi)
+          : appTraceTreeDataProvider.appTraceButton.label.label.match(
+              /start/gi
+            );
```
pwmb, line 1140: is there a change in the API?

brianignacio5: Yes, on [vscode November 2020](https://code.visualstudio.com/updates/v1_52#_markdown-tree-tooltip-api) the TreeItem API changed tooltip and label from string to `string | TreeItemLabel ` and `string | MarkdownString`

pwmb (approved):
LGTM, Thanks for this PR

src/rainmaker/view/item.ts
```diff
@@ -48,8 +48,9 @@ export class RMakerItem extends TreeItem {
   public set commandId(v: string) {
     this.command = {
       command: v,
-      title: this.label,
-      tooltip: this.tooltip,
+      title: typeof this.label === "string" ? this.label : this.label.label,
+      tooltip:
+        typeof this.tooltip === "string" ? this.tooltip : this.tooltip.value,
```
pwmb, line 53: This change is causing crash if user is not logged in already

brianignacio5: I'm not sure why would this crash the rainmaker tree view. Any additional information or related log ?

pwmb: if `this.tooltip == undefined` then it will crash because `.value` on `undefined`

PR #246, opened 2020-12-21
NVS Partition editor

brianignacio5:
Add NVS Partition editor to create CSV files with NVS architecture structure and generate binaries from these CSV files.

Based on [NVS Partition Generator Utility](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/storage/nvs_partition_gen.html) and #170


pwmb (approved):
Thanks for this feature LGTM!

PR #250, opened 2020-12-22
Full clean button and command

### LINKED ISSUE #247: [Feature Request]: Is there clean button? (VSC-509)
I don't find the "clean" button in the VS gui. I found out that with the terminal an typing idf.py fullclean is doing the same.

My problem cames that after a unsuccessful build I can not execute that command since I get this message:

image

If I close virual studio and open again then I can do a clean with idf.py fullclean.

brianignacio5:
Add `espIdf.fullClean` command to delete the build directory and status bar button.

Fix #247

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -431,8 +431,9 @@ export async function activate(context: vscode.ExtensionContext) {
       const buildDir = path.join(workspaceRoot.fsPath, "build");
       const buildDirExists = await utils.dirExistPromise(buildDir);
       if (!buildDirExists) {
-        Logger.infoNotify(`Directory ${buildDir} does not exists.`);
-        return;
+        return Logger.warnNotify(
+          `There is no build directory to clean, exiting!`
+        );
       }
       const cmakeCacheFile = path.join(buildDir, "CMakeCache.txt");
       const doesCmakeCacheExists = utils.canAccessFile(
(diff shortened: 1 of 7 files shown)


src/extension.ts
```diff
@@ -403,6 +404,29 @@ export async function activate(context: vscode.ExtensionContext) {
     });
   });
 
+  registerIDFCommand("espIdf.fullClean", () => {
+    PreCheck.perform([openFolderCheck], async () => {
+      const buildDir = path.join(workspaceRoot.fsPath, "build");
+      const buildDirExists = await utils.dirExistPromise(buildDir);
+      if (!buildDirExists) {
+        Logger.infoNotify(`Directory ${buildDir} does not exists.`);
```
pwmb, line 412: Maybe something like this `There is no build directory to clean, exiting!`. Also maybe its better to use warn than info

Also can use `return` in same line

src/extension.ts
```diff
@@ -403,6 +404,29 @@ export async function activate(context: vscode.ExtensionContext) {
...
+      const cmakeCacheFile = path.join(buildDir, "CMakeCache.txt");
+      const doesCmakeCacheExists = utils.canAccessFile(
+        cmakeCacheFile,
+        constants.R_OK
+      );
+      if (!doesCmakeCacheExists) {
+        Logger.infoNotify(
+          `Directory ${buildDir} doesn't seem to be a CMake build directory or is empty.`
+        );
+        return;
```
pwmb, line 424: same here!

pwmb (approved):
Left few suggestion

LGTM, thanks!

PR #255, opened 2020-12-30
Feature/doctor cmd

brianignacio5:
Add doctor command to create a report of user configuration. It will generate a `report.json` and `report.txt` (formatted output) files in EXTENSION_PATH and copy the content of `report.txt` to clipboard.

-- Extension configuration settings values.
-- Extension configuration settings values access.
-- ESP-IDF version, Python version, Pip version, Git version
-- Python packages and versions
-- Check ESP-IDF python packages
-- Check Extension python packages
-- Check Debug Adapter python packages

TO DO
- [ ] Add integration to Telemetry
- [x] Consider additional scenarios to test.


src/extension.ts
```diff
@@ -1219,6 +1220,10 @@ export async function activate(context: vscode.ExtensionContext) {
     });
   });
 
+  registerIDFCommand("espIdf.doctorCommand", async () => {
+    await generateConfigurationReport(context);
```
pwmb, line 1224: For my system this run very quickly but I still think to put this code inside `vscode.window.withProgress` API with `try/catch` to show error if something went wrong as well

pwmb (changes requested):
Some more points to the report would be great

- OS Details info (right now we only know win32 and arch)
- VS Code Version
- Extension Version
- Language of the vs code (en/etc)
- Check is debug config is proper or not (at least some basic keys)
- Check if c/c++ config is proper or not (like some some mandatory fields are set or not)

pwmb (approved):
Thanks @brianignacio5 LGTM!

pwmb:
You might need to add RU translations

PR #273, opened 2021-01-06
Add component manager UI Render (VSC-696)

pwmb:
Add component manager Render inside vscode also allow direct install button


src/component-manager/panel.ts
```diff
@@ -0,0 +1,143 @@
...
+      join(extensionPath, "media", "espressif_icon.png")
+    );
+    this.panel.webview.html = this.initWebView(url);
+  }
+  private onMessage(message: any) {
+    switch (message.message) {
+      case "install":
+        if (message.url) {
+          Logger.infoNotify(`Installing Component from ${message.url}`);
+        }
```
pwmb, line 95: @kumekay what's the command we need to run to install this component?

kumekay: @pwmb There is no such command in component manager since real dependency solving is only possible during run of CMake, 
what extension can do is to add a dependency to `idf_component.yml` file. This task was not implemented yet in component manager CLI, because there are no well-maintained yaml library for python with support for comments in yaml and we cannot reliably modify user-maintained yaml files.
So you can either implement it in extension itself, or I can add this feature.

If you want to actually download components after adding them to manifest you have to rerun CMake, however I think we shouldn't do it by default, because it may be slow.

pwmb: I think it will be better if we have a sub-command (in `idf.py` or relevant tool) to add entry to the manifest file and when user will build the project the dependency will be downloaded.

kumekay: Sure, it will be a part of `idf.py`,
Since it's possible to add dependencies not only to the main component of the project, but also for own components in `components` directory, full command will look like:

`idf.py add-dependency --component=main  example/cmp>=2.0.0 --add=expample/cmp2==1.0.0`

kumekay: We can only support adding dependencies to `main` component for projects or to components itself, if currently open code is idf component

kumekay: Going to work on the feature this week

pwmb: @brianignacio5 we can inject the command which is needed to add the deps once its available from IDF side, for now we are exposing the download url from the Component Manager UI, but I think we can expose more entities as required

kumekay: Sure, I'm sorry we changed the focus for last to sprints, so this feature is postponed a bit.

src/component-manager/panel.ts
```diff
@@ -0,0 +1,145 @@
...
+    this.panel.iconPath = Uri.file(
+      join(extensionPath, "media", "espressif_icon.png")
+    );
+    this.panel.webview.html = this.initWebView(url);
+  }
+  private onMessage(message: any) {
+    switch (message.message) {
+      case "install":
+        if (message.dependency) {
+          const component = message.component || "main";
```
brianignacio5, line 95: How does the Component Registry know to which component to add the dependency? 

Probably a future feature ?

brianignacio5:
LGTM !

src/component-manager/panel.ts
kumekay: For now, we only going to add dependencies to the main component (most popular use case), later we may add a way to select component. It will require changes for both the registry frontend and the vscode extension.

PR #279, opened 2021-01-06
Add environment variable support for Powershell Core for Linux and macOS

thezim:
PowerShell Core binary is called `pwsh` under Linux and macOS. This fix allows it to be detected properly.


brianignacio5 (approved):
Thanks for the contribution !

PR #319, opened 2021-02-19
Relocate CMakeLists Schema to root directory

### LINKED ISSUE #316: [Bug Report]: Failed reading CMakeLists.txt file when trying to use CMakeLists.txt Editor (VSC-570)
Latest VSCode-Extension built from source. (Issue was present in previous builds as well)
ArchLinux OS
IDF 4.4-dev

On any project, attempting to open the CMakeLists.txt Editor fails. The error is that the CMakeLists.txt file cannot be found even though it does exist. Both the Project and Component files do the same thing. If I step through the extension code by using the debug feature of VSCode Extension Host, it works fine. I think it may be a timing issue or some Async problem. I'm not that familiar with Type/JavaScript.
Here is what is logged:
``` bash
{"user":true,"message":"Failed reading /home/neal/esp/esp-adf/examples/get-started/play_mp3_control/CMakeLists.txt","stack":"TypeError: t is not iterable\n\tat /home/neal/.vscode/extensions/espressif.esp-idf-extension-0.6.1/dist/extension.js:8:242241\n\tat /home/neal/.vscode/extensions/espressif.esp-idf-extension-0.6.1/dist/extension.js:8:242496\n\tat Generator.next (<anonymous>)\n\tat s (/home/neal/.vscode/extensions/espressif.esp-idf-extension-0.6.1/dist/extension.js:8:241938)","level":"error","timestamp":"2021-02-18T23:08:40.465Z"}
```
Screenshot from 2021-02-18 17-08-42

brianignacio5:
Fix #316

Move CMakeListsSchema.json to root directory since it is needed on every CMakeLists.txt editor webview call.

Rename gulp_clean task to clean for clarity.


gulpfile.js
```diff
@@ -33,11 +33,13 @@ languagesDirs.forEach((langDir) => {
 function clean(done) {
   del([
     "dist/**",
+    "esp_idf_vsc_ext.log",
```
pwmb, line 36: Maybe its worth persisting the logs instead of deleting and if developer desire they can delete it themselves ?

PR #330, opened 2021-02-25
add and rm elements cmakeLists editor

brianignacio5:
Allow multiple entries of CMakeLists Element when reading and creating a CMakeLists.txt file.

Please take a look @meltdown03


meltdown03:
Thanks, it works now for multiple includes and I like the dropdown menu

PR #365, opened 2021-04-07
Doctor command testing

brianignacio5:
Add testing for doctor command.


gerekon (approved):
As far as I understand these tests are a kind of unit tests checking that functions used for report generation work properly.

I think that in the future we will also need to add e2e (end-to-end) doctor command test for some crucial scenarios. I mean to check the whole sequence:
0) Prepare specific env (IDF 4.0, 4.3, and maybe 3.3). For this first iteration we can do this just for the latest official IDF release (4.3 now)
1) Call doctor command. Do you know how to initiate command invocation from test script like it was done by user?
2) Check generated report.

PR #538, opened 2021-09-30
Enhance/partition table unit testing validation

radurentea:
- Add unit tests for partition table fields and CSV2JSON 
- Refactor duplicated function CSV2JSON
- Add validations and changed some for the partition table fields

### DIFF
--- a/src/test/suite/partitionTable.test.ts
+++ b/src/test/suite/partitionTable.test.ts
@@ -1,0 +1,501 @@
+import * as assert from "assert";
+import { isValidJSON, JSON2CSV } from "../../views/partition-table/util";
+import { CSV2JSON } from "../../views/partition-table/util";
+// You can import and use all API from the "vscode" module
+// as well as import your extension to test it
+import * as vscode from "vscode";
+import { stringify } from "querystring";
+import { PartitionTable } from "../../views/partition-table/store";
+import * as fse from "fs-extra";
+// import * as myExtension from "../extension";
+
+suite("PartitionTable Suite", () => {
+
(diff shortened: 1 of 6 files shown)


src/test/suite/partitionTable.test.ts
```diff
@@ -0,0 +1,501 @@
...
+import { stringify } from "querystring";
+import { PartitionTable } from "../../views/partition-table/store";
+import * as fse from "fs-extra";
+// import * as myExtension from "../extension";
+
+suite("PartitionTable Suite", () => {
+
+  test("CSV2JSON mockdata test", async () => {
+    const urlCSV = __dirname + "/../../../testFiles/test.csv";
+    const urlJSON = __dirname + "/../../../testFiles/test.json";
```
brianignacio5, line 16: I would recommend to use something like `resolve(__dirname, "..", "..", filename)` to avoid path separator issues across multiple OS.

src/test/suite/partitionTable.test.ts
```diff
@@ -0,0 +1,501 @@
+import * as assert from "assert";
+import { isValidJSON, JSON2CSV } from "../../views/partition-table/util";
+import { CSV2JSON } from "../../views/partition-table/util";
+// You can import and use all API from the "vscode" module
+// as well as import your extension to test it
+import * as vscode from "vscode";
+import { stringify } from "querystring";
+import { PartitionTable } from "../../views/partition-table/store";
+import * as fse from "fs-extra";
+// import * as myExtension from "../extension";
```
brianignacio5, line 10: Can remove commented lines.

PR #545, opened 2021-10-13
Enhance/nvs partition table unit tests

radurentea:
Unit tests are not included for u64 and i64 number types.


brianignacio5 (approved):
LGTM

PR #562, opened 2021-10-25
IDF_TARGET serialPort as workspace folder settings

### LINKED ISSUE #554: [Feature Request]:  Individual Port Control for separate open Projects/Windows (VSC-773)
When working on two ESP-IDF Projects in VSCode simultaneously, for example a Sender and Reciever nodes, then everytime I want to flash one of the boards, I have to keep switching the selected port in VSCode back and forth. Often I forget, and end up flashing a project to the wrong target. The problem is that when I set the port in one project, the same change automatically gets applied to the other project in another window. Strangely, this issue only seems to apply to serial port. All other settings seem to be easily individually customized for each project. Only the port seems to be a global setting (even though I'm using two different workspaces).

### LINKED ISSUE #559: [Bug Report]: Problems when switchting target (VSC-777)
**Describe the bug**
See bellow

**To Reproduce**
1.Copy the blink example, now there is no sdkconfig file in the project
2.Click the "ESP-IDF set espressif device target" button and select some target, now the sdkconfig file containing the right CONFIG_IDF_TARGET will be generated
3.Repeat 2 for any times, switching between differenct targets will work fine and sdkconfig file is refreshed with right CONFIG_IDF_TARGET value too
4.Click the "ESP-IDF SDK Configuration Editor(menuconfig)" button to open the SDK configuration editor tab, save it or not, close it or not.
5.Now when repeating 2 for the first time, sdkconfig won't be updated, when repeating 2 for more times, strange error message will show up.
6.The only way to reset is to manually delete sdkconfig file and build folder, even Full Clean won't work.

**Expected behavior**
At any time switching target should be fine and Full Clean if needed.

- Windows 10
- VS code version 1.61.2
- plug in version 1.2.0
- ESP-IDF version 4.3.1

brianignacio5:
Set serial port and IDF_TARGET as workspace folder specific settings.

Dispose conserver process before running `idf.py set-target`

Fix #559 

Fix #554

### DIFF
--- a/src/espIdf/serial/serialPort.ts
+++ b/src/espIdf/serial/serialPort.ts
@@ -96,13 +96,18 @@ export class SerialPort {
   }
 
   private async updatePortListStatus(l: string) {
-    const target = idfConf.readParameter("idf.saveScope");
-    await idfConf.writeParameter("idf.port", l, target);
+    const settingsSavedLocation = await idfConf.writeParameter(
+      "idf.port",
+      l,
+      vscode.ConfigurationTarget.WorkspaceFolder
+    );
     const portHasBeenSelectedMsg = this.locDic.localize(
       "serial.portHasBeenSelectedMessage",
       "Port has been updated to "
     );
-    Logger.infoNotify(portHasBeenSelectedMsg + l);
+    Logger.infoNotify(
+      `${portHasBeenSelectedMsg}${l} in ${settingsSavedLocation}`
+    );
   }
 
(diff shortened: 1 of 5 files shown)


crackwitz:
now, every time I use that command, it (1) doesn't remember the port and selects it in the menu (2) always asks me if I want to apply that to *this* workspace, which is the only one in the list.

can you please fix that so the command doesn't ask pointless questions over and over?

PR #566, opened 2021-11-03
WIP: VSC-294 DFU support

radurentea:


src/extension.ts
```diff
@@ -1188,6 +1188,7 @@ export async function activate(context: vscode.ExtensionContext) {
 
   registerIDFCommand("espIdf.createIdfTerminal", createIdfTerminal);
 
+  registerIDFCommand("espIdf.flashDFU", flash(false));
```
brianignacio5, line 1191: While trying to launch dev mode, I've received this error here:

```
ERROR in /Users/brainignacio/workspace/vscode-esp-idf-extension/src/extension.ts
    [tsl] ERROR in /Users/brainignacio/workspace/vscode-esp-idf-extension/src/extension.ts(1191,41)
          TS2345: Argument of type 'void' is not assignable to parameter of type '(...args: any[]) => any'.
```

This is because `flash(false)` is passing the result of the function not the function itself, which is the intended goal here.

src/extension.ts
```diff
@@ -2800,13 +2801,24 @@ const flash = () => {
...
         if (canFlash) {
-          await uartFlashCommand(
-            cancelToken,
-            flashBaudRate,
-            idfPathDir,
-            port,
-            workspaceRoot
-          );
+          if(isUart) {
+            await uartFlashCommand(
```
brianignacio5, line 2805: Seems that the difference here between if and else is an additional field. Why not passing `isUart` directly to `uartFlashCommand` method to reduce line and clarify the use. I would also rename `uartFlashCommand` to just `flashCommand` and `isUart` to an enum `[JTAG, DFU]` for clarity and type safety.

src/extension.ts
```diff
@@ -2866,7 +2878,7 @@ const buildFlashAndMonitor = async (runMonitor: boolean = true) => {
 async function selectFlashMethod(cancelToken) {
   let flashType = idfConf.readParameter("idf.flashType");
   if (!flashType) {
-    flashType = await vscode.window.showQuickPick(["JTAG", "UART"], {
+    flashType = await vscode.window.showQuickPick(["JTAG", "UART", "DFU"], {
```
brianignacio5, line 2881: Can reuse the enum types as string array to give the options to `showQuickPick` method.

src/flash/flashTask.ts
```diff
@@ -70,7 +70,7 @@ export class FlashTask {
     }
   }
 
-  public async flash() {
+  public async flash(isUart) {
```
brianignacio5, line 73: isUart could be rename to flashType and add the enum type here.

src/flash/flashTask.ts
```diff
@@ -98,8 +98,10 @@ export class FlashTask {
       powershellPath !== ""
     ) {
       flashExecution = await this._wslFlashExecution();
-    } else {
+    } else if (isUart) {
```
brianignacio5, line 101: Could use a switch with case from the flashType Enum here instead of multiple if else if

brianignacio5 (changes requested):
Thank you for your PR @radurentea 

I've left an issue I've found and code suggestions to improve readability and type safety.

src/utils.ts
```diff
@@ -959,6 +959,17 @@ export async function createNewComponent(
   );
 }
 
+export function selectedAdapterId(chip) {
```
brianignacio5, line 962: Maybe rename it to have more description about that is adapter id for DFU

src/flash/flashCmd.ts
```diff
@@ -79,5 +84,14 @@ export async function verifyCanFlash(
       new Error("NOT_SELECTED_BAUD_RATE")
     );
   }
+  const selectedFlashType = idfConf.readParameter("idf.flashType");
+  if (selectedFlashType === "DFU") {
+    if (!(await listDfuDevices())) {
```
brianignacio5, line 89: What kind of information you get from `dfu-util --list` ? Is it possible to validate that the device defined for current target (esp32, esp32s2) is listed in `listDfuDevices()` output ?

brianignacio5 (changes requested):
Thank you for the changes.

Left some comments for improvements in validation.

src/flash/flashCmd.ts
radurentea: This is an example of what I receive from `dfu-util --list`  I receive for esp32s2:
`Found Runtime: [303a:0002] ver=0723, devnum=1, cfg=1, intf=2, path="0-1", alt=0, name="UNKNOWN", serial="0"`

I don't think I can validate it

package.json
```diff
@@ -1239,5 +1245,8 @@
     "insight": {
       "key": "912cbb4e-e606-4c4f-9a00-1731acb15117"
     }
-  }
+  },
```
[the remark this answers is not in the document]
radurentea: Thanks! I've removed this section in my last commit

src/build/buildTask.ts
```diff
@@ -140,4 +163,34 @@ export class BuildTask {
       showTaskOutput
     );
   }
+
+  public async buildDfu() {
+    this.building(true);
+    const modifiedEnv = appendIdfAndToolsToPath();
+    await ensureDir(this.curWorkspace);
+
```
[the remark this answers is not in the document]
radurentea: Thanks! I've added validation for it in my last commit.

brianignacio5:
We should also add target validation. For example, if the user tries to use DFU but the current target is `esp32` it will still tries to execute these:

1) Say a user just do `Select flash method and flash`, choose DFU and it will try to flash:

The task `dfu-util -d 303a:undefined -D /Users/myUser/workspace/blink/build/dfu.bin` will try to execute. There should be IDF_TARGET  (esp32s2, esp32s3) validation and that `dfu.bin` exists.

2) Say the user do `Build your project` which will create `dfu.bin`

The task `/Users/myUser/esp/esp-idf/tools/mkdfu.py write -o /Users/myUser/workspace/blink/build/dfu.bin --json /Users/myUser/workspace/blink/build/flasher_args.json --pid undefined`.

IDF_TARGET (esp32s2, esp32s3) validate should exist before.

src/flash/flashTask.ts
```diff
@@ -138,6 +148,15 @@ export class FlashTask {
     return new vscode.ProcessExecution(pythonBinPath, flasherArgs, options);
   }
 
+  public _dfuFlashing() {
+    this.flashing(true);
+    return new vscode.ShellExecution(
```
brianignacio5, line 153: This method of flashing doesn't consider the case where multiple chips are present and how to choose between devices. As shown in [DFU documentation](https://docs.espressif.com/projects/esp-idf/en/latest/esp32s2/api-guides/dfu.html), you should use the `--path` argument and provide the user with the chance to choose which device to use.

chipweinberger:
Found this page from the VSC 1.4 changelog, what is this? Support for flashing & JTAG over the native USB found on the S2 and S3?

chipweinberger:
This forum post might be more useful that the docs: 
https://www.esp32.com/viewtopic.php?f=12&p=94166

PR #645, opened 2022-02-13
More CCache fixes

meltdown03:
- When adding components form the Component Registry, run `reconfigure` also so it uses the components on the next build
- Use `--ccache' arg, if enabled, when adding components and setting target board


brianignacio5:
Needed to refactor a bit, added these changes to master directly

PR #668, opened 2022-03-14
add custom build folder by conf setting

brianignacio5:
Add `idf.buildDirectoryName` to specify a custom build folder for the extension commands. By default is `build`

### DIFF
--- a/docs/SETTINGS.md
+++ b/docs/SETTINGS.md
@@ -14,6 +14,7 @@ These are the configuration settings that ESP-IDF extension contributes to your 
 
 | Setting ID                      | Description                                                                   |
 | ------------------------------- | ----------------------------------------------------------------------------- |
+| `idf.buildDirectoryName`        | Custom build directory name for extension commands. (Default: build)          |
 | `idf.cmakeCompilerArgs`         | Arguments for CMake compilation task                                          |
 | `idf.customExtraPaths`          | Paths to be appended to \$PATH                                                |
 | `idf.customExtraVars`           | Variables to be added to system environment variables                         |
--- a/i18n/en/package.i18n.json
+++ b/i18n/en/package.i18n.json
@@ -87,6 +87,7 @@
   "param.preFlashTask": "Pre flash custom task",
   "param.postFlashTask": "Post flash custom task",
   "param.customTask": "Custom task",
+  "param.buildDirectoryName": "Name of CMake build directory",
   "view.components.name": "Project Components",
   "configuration.title": "ESP-IDF",
   "espIdf.apptrace.archive.refresh.title": "Refresh Trace Archive List",
--- a/i18n/es/package.i18n.json
+++ b/i18n/es/package.i18n.json
@@ -90,6 +90,7 @@
   "param.preFlashTask": "Pre build tarea personalizada",
   "param.postFlashTask": "Post build tarea personalizada",
   "param.customTask": "Tarea personalizada",
+  "param.buildDirectoryName": "Nombre del directorio de construcción de CMake",
   "view.components.name": "Componentes de proyecto",
   "configuration.title": "ESP-IDF",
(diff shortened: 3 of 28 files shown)


project-x51:
So using this feature I could make the code build to a RAM disk?

project-x51:
I was able to make it build to a different folder, but not to another drive. On Windows 11.  
NOTE: I have built to a RAM disk before using the idf.py command but not from VSCode. The build to the RAM disk was about 3 x as fast.

brianignacio5:
This build folder is still relative to the location of the project directory. Resulting build directory is `${workspaceFolder}/${idf.buildDirectoryName}`. We could modify it to specify a full build path if this is your intended purpose here.

project-x51:
Yeah, that would be awesome. Building to a RAM disk is faster and it doesn't wear my SSD writing files I don't need to keep.

project-x51:
The absolute coolest would be if I could specify the build directory anywhere and then copy the distribution files to a different folder after the build. This way I could build to a RAM disk and keep the .bin and .elf files under the project in a dist folder.

brianignacio5:
Could you try this [vsix installer](https://nightly.link/espressif/vscode-esp-idf-extension/actions/artifacts/345066569.zip) ?

[how to Install extension from vsix](https://code.visualstudio.com/docs/editor/extension-marketplace#_install-from-a-vsix)

project-x51:
Sorry, I just got a break in things to try this. 

It doesn't appear to work and always builds to the workspace build folder no matter what I put in the setting. The vsix installer says it installed the extension. It still has 1.5.0 as the version so ??

`{
    "idf.adapterTargetName": "esp32s3",
    "idf.portWin": "COM4",
    "idf.flashType": "UART",
    "files.associations": {
        "semphr.h": "c"
    },
    "idf.openOcdConfigs": [
        "board/esp32s3-bridge.cfg"
    ],
    "idf.buildDirectoryName": "build 1"
}`

or 

`{
    "idf.adapterTargetName": "esp32s3",
    "idf.portWin": "COM4",
    "idf.flashType": "UART",
    "files.associations": {
        "semphr.h": "c"
    },
    "idf.openOcdConfigs": [
        "board/esp32s3-bridge.cfg"
    ],
    "idf.buildDirectoryName": "c:/temp/build 1"
}`

Both build into the 'Build' folder, even if I make the 'build 1' folder manually.

brianignacio5:
You still can't use spaces in the build path unfortunately... Did you tried without spaces ?

brianignacio5:
Maybe there is a project settings.json ` idf.buildDirectoryName` value that is overriding the setting as specified in [multiple projects](https://github.com/espressif/vscode-esp-idf-extension/blob/master/docs/MULTI_PROJECTS.md) ?

project-x51:
I don't think so I can't build to anywhere else even with only the Hello_World example project open in VSCode

brianignacio5:
This MR is just to change the name of the build folder within your current project. 

Let's say `C:\temp\hello_world\build` when idf.buildDirectoryName is default and `C:\temp\hello_world\custom_build` when  idf.buildDirectoryName is `custom_build`.

The MR with the custom absolute build path is [here](https://github.com/espressif/vscode-esp-idf-extension/pull/794)

project-x51:
I defiantly have the change in this MR. If I type "idf.buildDir... " then it autocompletes with "idf.buildDirectoryName": "${workspaceFolder}/build". This change was made in your comit.

project-x51:
There are a lot of changes in the commit. I'm not up to speed on everything in there but I notice you have two names for the build directory. I'm guessing there is an inconsistency related to that. 

example the change in **src/coverage/coverageService.ts**. The build path is removed in the join statement but the parameter below remains the same (**buildDirName**). Elsewhere such as in **src/build/buildCmd.ts** the idf.buildDirectoryName is mapped to the variable **buildDirPath**

project-x51:
I'll post in the MR and continue our conversation there.

PR #685, opened 2022-03-28
Add unit tests for Test Coverage

radurentea:


brianignacio5:
Could you simplify the required files for these testing ?

`gcovr` only needs the `.gcda` and `.gcno` files in the project. So having the `.c` files should be enough to run gcovr and get the output.

You could test the html against a given expected output in `buildHtml` function. The `generateCoverageForEditors` function should also be more descriptive,

brianignacio5 (changes requested):
The ` test("generateCoverageForEditors")` is not testing anything.

Should open one of the test project files and test that `result` array contain an editor with same filePath as one of the file from the test project.

PR #716, opened 2022-05-16
Add all output channels to unified log file

radurentea:


src/logger/outputChannel.ts
```diff
@@ -22,14 +22,24 @@ export class OutputChannel {
...
-  public static append(message: string) {
+  public static append(message: string, name?: string) {
     OutputChannel.checkInitialized();
-    OutputChannel.instance.append(message);
+    if(name) {
+      OutputChannel.instance.appendLine(`[${name}]`);
+      OutputChannel.instance.append(message);
+    } else {
+      OutputChannel.instance.append(message);
+    }
```
brianignacio5, line 42: Could be simplified with 
```ts
if (name) {
   OutputChannel.instance.append(`[${name}]`);
}
OutputChannel.instance.append(message);
```

src/logger/outputChannel.ts
```diff
@@ -22,14 +22,24 @@ export class OutputChannel {
...
-  public static appendLine(message: string) {
+  public static appendLine(message: string, name?: string) {
     OutputChannel.checkInitialized();
-    OutputChannel.instance.appendLine(message);
+    if(name) {
+      OutputChannel.instance.appendLine(`[${name}]`);
+      OutputChannel.instance.appendLine(message);
+    } else {
+      OutputChannel.instance.appendLine(message);
+    }
```
brianignacio5, line 32: Same simplification here.

brianignacio5 (changes requested):
Any reason for not implementing those OutputChannel `name?: string` tags here if the function is already present? Would be harder to track in a different PR I think but up to you.

src/logger/outputChannel.ts
radurentea: Thanks!

src/logger/outputChannel.ts
radurentea: Thank you, I've changed both.

brianignacio5 (approved):
LGTM

PR #724, opened 2022-05-26
add built in esp32s3 default board

### LINKED ISSUE #710: Select openOCD board configuration missing esp32-s3 usb-jtag option.... this was painful to get working (VSC-897)
esp-box dev kit here.

Ran through the installation steps, was able to program via openocd but kept getting invalid target errors, I think because I was using the esp32c3 openocd configuration as that was the closest one to the esp32s3 config with built in usb-jtag.

Started going through the bug reporting process and pulling variables and then, after verifying that the appropriate board file existed changed these entries from:

```
    "idf.openOcdConfigs": [
        "board/esp32c3-builtin.cfg"
    ],
    "idf.adapterTargetName": "esp32c3",
```

to:

```
    "idf.openOcdConfigs": [
        "board/esp32s3-builtin.cfg"
    ],
    "idf.adapterTargetName": "esp32s3",
```

then restarted vscode and its working, so I think this is what fixed it.

Is there a reason there is no esp32-s3 usb-jtag option for the openocd board config?

brianignacio5:
Fix #710

### DIFF
--- a/src/espIdf/openOcd/boardConfiguration.ts
+++ b/src/espIdf/openOcd/boardConfiguration.ts
@@ -48,6 +48,12 @@ export const defaultBoards = [
     configFiles: ["interface/ftdi/esp32_devkitj_v1.cfg", "target/esp32s3.cfg"],
   } as IdfBoard,
   {
+    name: "ESP32-S3 chip (via builtin USB-JTAG)",
+    description: "ESP32-S3 used with ESP-PROG board",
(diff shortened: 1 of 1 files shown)


Freedom1986:
need to fix:
description: "ESP32-S3 used with ESP-PROG board",
To
description: "ESP32-S3 debugging via builtin USB-JTAG",

PR #726, opened 2022-05-31
fix device partition explorer issues with custom offset

boarchuz:
- Cancel `populatePartitionItems` if custom offset input box loses focus
 - Fix hexadecimal input for custom offset: '0x' prefix was prevented by regex validation but is required by esptool.py and gen_esp32part.py
 - Pass custom offset to gen_esp32part.py to avoid misleading errors caused by using default partition table offset instead


brianignacio5:
Thanks for the contribution @boarchuz

PR #871, opened 2023-01-10
ESP matter device path (VSC-1032)

brianignacio5:
## Description

Fix the `ESP_MATTER_DEVICE_PATH` by adding the `Set ESP-MATTER Device Path (ESP_MATTER_DEVICE_PATH)` command to set the value to the `idf.espMatterDevicePath` configuration setting which is used by the extension with fallback in  system `ESP_MATTER_DEVICE_PATH` environment variable.

## Type of change

## How has this been tested?

Test by manually testing the feature

**Test Configuration**:
* ESP-IDF Version: 5.1

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting
- [x] Added Documentation
- [ ] Added Unit Test


brianignacio5:
Could you please take a look at this PR @Diegorro98  ?

Diegorro98:
@brianignacio5 I can't test it right now, but LGTM.

Diegorro98:
I have a question, if this pull request is to solve this problem: https://github.com/espressif/esp-matter/issues/104#issuecomment-1376074119, is it really necessary to define `ESP_MATTER_DEVICE_PATH`? I mean, I've seen that the only use case of `ESP_MATTER_DEVICE_PATH` is at the esp-matter examples. Maybe a better approach is forget about defining `ESP_MATTER_DEVICE_PATH` an let the examples define it using the IDF_TARGET in CMakeLists.txt (i.e. light example [CMakeLists.txt](https://github.com/espressif/esp-matter/blob/716353fecbdc601112dc5be2a44da1a6cfe7096c/examples/light/CMakeLists.txt#L9-L21))

eos1d3:
@brianignacio5 
I created my own light example and created my own device inside the project. The pre-defined ESP_MATTER_DEVICE_PATH is causing problem for my project.

If I use terminal to build, there is no problem at all for my own project. But using VSCode, ESP_MATTER_DEVICE_PATH is already defined and not points to my own device path.

So I also suggest ESP_MATTER_DEVICE_PATH should **NOT** be defined.

brianignacio5:
I removed setting the `ESP_MATTER_DEVICE_PATH` in the extension but kept the command to add it as environment variable with the `Set ESP-MATTER Device Path (ESP_MATTER_DEVICE_PATH)`.

PTAL at the artifact here @eos1d3 @Diegorro98

Diegorro98:
LGTM, I've tried with and without defining `ESP_MATTER_DEVICE_PATH` and both do its job perfectly

PR #910, opened 2023-03-17
update non root container dockerfile

### LINKED ISSUE #883: Non root containers (VSC-1042)
**Is your feature request related to a problem? Please describe.**
When working inside a docker dev container, you are effectively using a root account in the container. All files created or built will be owned by root. Even if the security implications are not that high, it's a hassle when we use the repo outside the jail, as the files will be owned by root outside the container.

I like to have vscode create a user inside the jail, with the same UID and GID as the external user. Also, mount the workspace in exactly the same path as inside, so the full path you copy from vscode is the same externally making it easy to copy and paste to an external terminal.

**Additional context**

brianignacio5:
## Description

Define non root user in Dockerfile and update template devcontainer.json to latest schema for settings and extensions.

Fixes #883

## Type of change

## How has this been tested?

Manual testing opening a project in a docker container.

**Test Configuration**:
* ESP-IDF Version: 4.4.2

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/templates/.devcontainer/Dockerfile
+++ b/templates/.devcontainer/Dockerfile
@@ -1,6 +1,9 @@
 FROM espressif/idf
 
 ARG DEBIAN_FRONTEND=nointeractive
+ARG CONTAINER_USER=esp
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
 
 RUN apt-get update \
   && apt install -y -q \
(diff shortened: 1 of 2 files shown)


radurentea (approved):
LGTM

PR #945, opened 2023-04-19
trim nvs csv columns

### LINKED ISSUE #937: NVS Partition Editor doesn't read CSV files with CRLF endings under Windows (VSC-1088)
### OS

Windows

### Operating System version

Windows 10

1.76.2

### ESP-IDF version

n/a

### Python version

n/a

### Doctor command output

n/a

### Extension

_No response_

### Description

The NVS Partition Table Editor (used to edit the csv files that can be turned into a binary partition to flash to an NVS partition) does not accept a file with CRLF line endings on Windows when the file contains a line with an integer encoding (e.g. `u32`). The file will open, but the UI will not be populated with the contents of the file.

Splitting the CSV file into lines happens here: https://github.com/espressif/vscode-esp-idf-extension/blob/af9f3776db92a2eaadfab44b892757a00b205a5c/src/views/nvs-partition-table/util.ts#L152

Notably, I do not have issues with the _Partition Table_ editor which does the same split operation here: https://github.com/espressif/vscode-esp-idf-extension/blob/af9f3776db92a2eaadfab44b892757a00b205a5c/src/views/partition-table/util.ts#L134

I think this one works because each cell value is trimmed before being processed, which removes the `\r` characters. Meanwhile, the NVS editor lets the `\r` character through and it stays attached to the last cell (`value`). For numerical values, this breaks with this regex check:
https://github.com/espressif/vscode-esp-idf-extension/blob/af9f3776db92a2eaadfab44b892757a00b205a5c/src/views/nvs-partition-table/util.ts#L112

I'm not sure why `os.EOL` is returning `\n` and not `\r\n` under Windows. Perhaps extensions run in a web view where `os` is polyfilled, and the implementation always uses `\n`. Either way it seems like an easy fix to also trim every cell in the NVS Partition Table Editor. This may be an issue for `string` type columns that contain intentional whitespace, so perhaps instead we could just sanitise the carriage returns only. Alternatively, we could find a mechanism to choose the correct line ending to split on instead of `os.EOL` which appears to be incorrect under Windows.

### Debug Message

```plain
n/a
```

### Other Steps to Reproduce

On Windows, create an NVS partition CSV file using CRLF line endings and include a few lines with `u32` encoding type. Alternatively, generate one with the NVS Partition Table Editor and manually change the line endings from LF to CRLF.

Opening such a file in the NVS Partition Editor leads to an empty UI, and changing the CRLF endings to LF populates the UI correctly.

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

brianignacio5:
## Description

Trim columns in NVS partition table editor

Fixes #937

## Type of change

1. Right click on a NVS partition table csv with CRLF line ending
2. click `ESP-IDF: Open NVS Partition Editor` command
4. Observe results. The table values should shown in the editor.

- Expected behaviour: The table values should shown in the editor.

- Expected output: The table values should shown in the editor.

## How has this been tested?

Manual test using csv with CRLF end of lines characters.

**Test Configuration**:
* ESP-IDF Version: 5.1

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting
- [x] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/src/test/suite/nvsPartitionTable.test.ts
+++ b/src/test/suite/nvsPartitionTable.test.ts
@@ -8,11 +8,6 @@ import {
   minValues,
   maxValues
 } from "../../views/nvs-partition-table/util";
-// You can import and use all API from the "vscode" module
-// as well as import your extension to test it
-import * as vscode from "vscode";
-import { stringify } from "querystring";
-import { NvsPartitionTable } from "../../views/nvs-partition-table/store/index";
 import * as fse from "fs-extra";
 import { resolve } from "path";
 
--- a/src/views/nvs-partition-table/util.ts
+++ b/src/views/nvs-partition-table/util.ts
@@ -129,10 +129,8 @@ export function isInValidRow(row: NvsPartitionTable.IRow): string {
   return;
 }
 
-export const expectedHeader = "key,type,encoding,value";
-
 export function JSON2CSV(rows: NvsPartitionTable.IRow[]) {
-  let csv = expectedHeader + EOL;
+  let csv = `key, type, encoding, value ${EOL}`;
   for (const row of rows) {
     if (
       row.key === "" &&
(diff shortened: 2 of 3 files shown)


radurentea:
LGTM

PR #1007, opened 2023-07-08
Update install.md (VSC-1139)

fgnievinski:
minor documentation update to avoid common mistake


brianignacio5:
Could you provide more details about what common mistakes you would avoid with this note ?

fgnievinski:
> Could you provide more details about what common mistakes you would avoid with this note ?

If one tries to configure an installation made several months ago, it's better to pick a more recent version.

brianignacio5:
Closing this PR because it doesn't contribute to fix any issue. If you can specify what issues are present when using an older version of ESP-IDF installed it would be good to fix those issues instead.

PR #1171, opened 2024-03-26
feat: avoid adding `saveDefSdkconfig` menu item to every files in explorer

ArcticLampyrid:
## Description

Avoid adding `saveDefSdkconfig` menu item to every files in explorer

## Type of change
New feature (non-breaking change which adds functionality)

## Expected behaviour
`espIdf.saveDefSdkconfig` should be only added to ESP-IDF-related project files.
And it should not be ordered at first, for it is not used often.

### DIFF
--- a/package.json
+++ b/package.json
@@ -282,8 +282,9 @@
           "group": "navigation"
         },
         {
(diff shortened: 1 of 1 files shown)


brianignacio5:
LGTM thanks for the changes !

PR #1228, opened 2024-06-26
Add validation for existing setup path

radurentea:
## Description

Added validation for esp-idf path, when user selects "Find ESP-IDF in your system" while configuring the extension.

Jira: [VSC-1174](]https://jira.espressif.com:8443/browse/VSC-1174)

## Type of change
- Enhancement

- Configure extension:
image
- Choose "Select ESP-IDF version:

1. Try to do it with a valid esp-idf folder (it should work)
2. Try to do it with an invalid esp-idf folder (error should appear and install button should be disabled)
image

## How has this been tested?
As described above

**Test Configuration**:
* ESP-IDF Version: 5.2.2

## Checklist
- [ ] PR Self Reviewed
- [x] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/src/setup/SetupPanel.ts
+++ b/src/setup/SetupPanel.ts
@@ -46,7 +46,13 @@ import { createPyReqs } from "./pyReqsInstallStep";
 import { downloadIdfTools } from "./toolsDownloadStep";
 import { installIdfGit, installIdfPython } from "./embedGitPy";
 import { getOpenOcdRules } from "./addOpenOcdRules";
-import { checkSpacesInPath, getEspIdfFromCMake, canAccessFile, execChildProcess, compareVersion } from "../utils";
+import {
+  checkSpacesInPath,
+  getEspIdfFromCMake,
+  canAccessFile,
+  execChildProcess,
+  compareVersion,
+} from "../utils";
 import { useIdfSetupSettings } from "./setupValidation/espIdfSetup";
 import { clearPreviousIdfSetups } from "./existingIdfSetups";
 
@@ -328,20 +334,27 @@ export class SetupPanel {
         case "canAccessFile":
           if (message.pathIdfPy) {
             const fileExists = await canAccessFile(message.pathIdfPy);
-            if(!fileExists) {
+            if (!fileExists) {
               this.panel.webview.postMessage({
                 command: "canAccessFileResponse",
                 path: message.path,
                 exists: fileExists,
               });
             } else {
-              const versionEspIdf = await getEspIdfFromCMake(message.path);
-              const noWhiteSpaceSupport = compareVersion("5.0", versionEspIdf);
+              let versionEspIdf = message.currentVersion;
+              if (!versionEspIdf) {
+                versionEspIdf = await getEspIdfFromCMake(message.path);
+              }
+              // compareVersion returns a negative value if versionEspIdf is less than "5.0"
+              const noWhiteSpaceSupport =
+                compareVersion(versionEspIdf, "5.0") < 0;
+              const hasWhitespace = /\s/.test(message.path);
               this.panel.webview.postMessage({
                 command: "canAccessFileResponse",
                 path: message.path,
                 exists: fileExists,
-                noWhiteSpaceSupport
+                noWhiteSpaceSupport,
+                hasWhitespace,
               });
             }
           }
@@ -355,7 +368,8 @@ export class SetupPanel {
   }
 
   setupErrHandler(error: Error) {
-    const errMsg = error && error.message ? error.message : "Error during ESP-IDF setup";
+    const errMsg =
+      error && error.message ? error.message : "Error during ESP-IDF setup";
     if (
       errMsg.indexOf("ERROR_EXISTING_ESP_IDF") !== -1 ||
       errMsg.indexOf("IDF_PATH_WITH_SPACES") !== -1 ||
--- a/src/setup/SetupPanel.ts
+++ b/src/setup/SetupPanel.ts
@@ -342,8 +342,13 @@ export class SetupPanel {
                 exists: fileExists,
               });
             } else {
-              let versionEspIdf = message.currentVersion;
-              if (!versionEspIdf) {
+              let versionEspIdf;
+              if (
+                message.currentVersion &&
+                typeof message.currentVersion === "string"
+              ) {
+                versionEspIdf = message.currentVersion;
+              } else {
                 versionEspIdf = await getEspIdfFromCMake(message.path);
               }
               // compareVersion returns a negative value if versionEspIdf is less than "5.0"
--- a/src/setup/SetupPanel.ts
+++ b/src/setup/SetupPanel.ts
@@ -46,7 +46,7 @@ import { createPyReqs } from "./pyReqsInstallStep";
 import { downloadIdfTools } from "./toolsDownloadStep";
 import { installIdfGit, installIdfPython } from "./embedGitPy";
 import { getOpenOcdRules } from "./addOpenOcdRules";
-import { checkSpacesInPath, getEspIdfFromCMake } from "../utils";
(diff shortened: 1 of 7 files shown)


brianignacio5 (approved):
Few comments but otherwise LGTM

src/views/setup/Install.vue
```diff
@@ -101,64 +160,6 @@ function setToolsFolder(newToolsPath: string) {
 }
 </script>
 
-<template>
-  <div id="install">
-    <div class="notification is-danger error-message" v-if="espIdfErrorStatus">
-      <p>{{ espIdfErrorStatus }}</p>
-      <div class="icon is-large is-size-4" @click="setEspIdfErrorStatus">
-        <IconClose />
```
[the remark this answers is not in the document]
radurentea: I've moved it to the top cause I think it's the standard practice and I prefer it. (template, script, style)
I also noticed that we don't have a consistent way of doing it in the application, but I can change it back to how it was.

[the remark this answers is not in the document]
brianignacio5: https://vuejs.org/style-guide/rules-recommended.html#single-file-component-top-level-element-order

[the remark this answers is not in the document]
radurentea: Thank you, I wasn't aware the script,template,style is also considered good practice.
I've checked now the extension and it actually seems we are consistent with script, template, style so I will revert the changes to this PR. 
I think I've looked at a file that didn't actually had a script tag when I've said that we are not consistent with it.

src/setup/SetupPanel.ts
```diff
@@ -83,6 +84,14 @@ export class SetupPanel {
   private readonly panel: WebviewPanel;
   private disposables: Disposable[] = [];
 
+  private async checkFileExists(filePath: string): Promise<boolean> {
```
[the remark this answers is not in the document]
radurentea: Thanks! I've changed it with my last commit.

radurentea:
This should not be tested yet. New improvements were requested and I'm currently working on implementing them.

radurentea:
All changes are done, can be tested

AndriiFilippov:
@radurentea hi !

Tested under:
OS - Windows 10 

tested with multiple esp-idf folders, do see validation ✔️ 

LGTM 👍

radurentea:
@AndriiFilippov Thank you, I've fixed the issue. When you have time please test it again

radurentea:
image
Fixed and I've also stylized a bit also the following bug found by Andrii on the configuration page, where we were displaying the text wrongly.

AndriiFilippov:
@radurentea hi ! 

good 👍 

validation should works only for existing ESP-IDF folder.
Download ESP-IDF case: 

image

AndriiFilippov:
@radurentea hi ! 

everything works ✔️ 
LGTM 👍 
but it is still uncomfortable to get validation just on unfocus. Are there any chances to somehow maneuver around this situation?

radurentea:
Hi Andrii, we've discussed this before the implementation and we agreed on this. We can't make it real time because there will be to many calls

AndriiFilippov (approved):

radurentea:
Hi @brianignacio5, PTAL again

src/views/setup/components/selectEspIdf.vue
```diff
@@ -108,6 +112,16 @@ watchEffect(() => {
     store.whiteSpaceErrorIDFContainer = "";
   }
 });
+
+watch(selectedEspIdfVersion, (newVal) => {
+  clearIDfErrorStatus();
+  store.clearIdfPathError();
+  if (newVal.filename === "manual") {
```
brianignacio5, line 119: Seems to me that ` validatePathOnBlur` and this watch are doing the same thing. Could we remove the code of one of these 2 ?

src/views/setup/store.ts
```diff
@@ -97,6 +98,83 @@ export const useSetupStore = defineStore("setup", () => {
...
+    idfPathError.value = error;
+    isInstallButtonDisabled.value = !!error;
+  }
+
+  function validateEspIdfPath(path: string) {
+    clearIdfPathError();
+    vscode.postMessage({
+      command: "canAccessFile",
+      path,
+      pathIdfPy: `${path}/tools/idf.py`,
```
brianignacio5, line 119: Does `pathIdfPy` needs to be set here ? I think it could be generated in the vscode panel side. Why ? Well What about Windows-style paths `\\` ?

src/views/setup/store.ts
```diff
@@ -97,6 +98,83 @@ export const useSetupStore = defineStore("setup", () => {
   let toolsFolder: Ref<string> = ref("");
   let toolsResults: Ref<IEspIdfTool[]> = ref([]);
   let extensionVersion: Ref<string> = ref("");
+  let idfPathError: Ref<string> = ref("");
+  let isInstallButtonDisabled: Ref<boolean> = ref(false);
+
+  function clearIdfPathError() {
```
brianignacio5, line 104: Isn't this function just `setIdfPathError("");` We could remove this function I think. You also use `setIdfPathError("");` for canAccessFileResponse window event listener.

src/setup/SetupPanel.ts
```diff
@@ -325,6 +331,34 @@ export class SetupPanel {
         case "exploreComponents":
           await commands.executeCommand("esp.component-manager.ui.show");
           break;
+        case "canAccessFile":
+          if (message.pathIdfPy) {
+            const fileExists = await canAccessFile(message.pathIdfPy);
```
brianignacio5, line 336: I think `message.pathIdfPy` can be generated here instead of generating it in the Vue UI.

brianignacio5 (changes requested):
A few code duplication changes required.

On MacOS, I see the following error after pressing Express install option and the UI is empty so can't continue:
Screenshot 2024-07-09 at 16 51 24

src/views/setup/components/selectEspIdf.vue
radurentea: I couldn't find a solution in which both are not needed. 

The reason is the `@blur` event, which doesn't trigger correctly if I would use store.validateEspIdfPath, I need the verification of `selectedEspIdfVersion.value.filename === "manual"`

radurentea:
> A few code duplication changes required.
> 
> On MacOS, I see the following error after pressing Express install option and the UI is empty so can't continue: Screenshot 2024-07-09 at 16 51 24

Hi @brianignacio5, please take another look with the new updates. I've tested on Mac and I couldn't not reproduce the error.

src/setup/SetupPanel.ts
```diff
@@ -325,6 +331,35 @@ export class SetupPanel {
...
+            const pathIdfPy = path.join(message.path, "tools", "idf.py");
+            const fileExists = await canAccessFile(pathIdfPy);
+            if (!fileExists) {
+              this.panel.webview.postMessage({
+                command: "canAccessFileResponse",
+                path: message.path,
+                exists: fileExists,
+              });
+            } else {
+              let versionEspIdf = message.currentVersion;
```
brianignacio5, line 345: Maybe we should validate that `message.currentVersion` does exist ?

brianignacio5 (approved):
Tested on MacOS and the previous issue doesn't happen anymore.

LGTM

src/setup/SetupPanel.ts
radurentea: I agree! I've added the validation.

radurentea:
Hi @AndriiFilippov, please test when you have a moment.

AndriiFilippov:
@radurentea hi !

Tested under:
OS - Windows 11 / MacOS 

LGTM 👍

AndriiFilippov (approved):

PR #1229, opened 2024-06-28
update zh-cn (VSC-1416)

Tinyu-Zhao:
I'm a Chinese developer and some of the translations are too hard to understand, if you don't mind, I can help you to improve them.

flash in these three places as a verb, in the Chinese translation for the burning device is easier to understand.
Save translates to "节省"

Full Clean directly translated bit "完全清洁" makes it difficult to understand, the action is to remove the contents of the build, the Chinese translation is "清除构建" more understandable. 

There are also nouns such as Rainmaker which translates to "造雨者."


brianignacio5:
Hi @Tinyu-Zhao Thank you for your feedback. We are currently preparing more formal translation for the extension commands and documentation.

As my Chinese is limited I will ask for my colleagues for review and we will have our documentation team for better docs.

Apologies and thank you.

PR #1252, opened 2024-07-17
fix support for multiple sdkconfig files

brianignacio5:
## Description

Fix support of multiple sdkconfig when using Project Configuration and multiple profiles.

Fixes #1202
Fixes #1265
Fixes #1283
Fixes #1299

## Type of change

Please delete options that are not relevant.
- Bug fix (non-breaking change which fixes an issue)

## Steps to test this pull request

Using a project with multiple configurations like https://github.com/espressif/esp-idf/tree/master/examples/build_system/cmake/multi_config run the `ESP-IDF: Open Project Configuration` and configure 2 profiles with different SDK Configuration Editor, resulting in a `esp_idf_project_configuration.json` like this:

```json
{
  "gen1": {
    "build": {
      "compileArgs": [],
      "ninjaArgs": [],
      "buildDirectoryPath": "${workspaceFolder}/build_gen1",
      "sdkconfigDefaults": [
        "sdkconfig.prod_common",
        "sdkconfig.prod1"
      ],
      "sdkconfigFilePath": "${workspaceFolder}/sdkconfig1"
    },
    "env": {},
    "flashBaudRate": "",
    "idfTarget": "",
    "monitorBaudRate": "",
    "openOCD": {
      "debugLevel": 0,
      "configs": [],
      "args": []
    },
    "tasks": {
      "preBuild": "",
      "preFlash": "",
      "postBuild": "",
      "postFlash": ""
    }
  },
  "gen2": {
    "build": {
      "compileArgs": [],
      "ninjaArgs": [],
      "buildDirectoryPath": "${workspaceFolder}/build_gen2",
      "sdkconfigDefaults": [
        "sdkconfig.prod_common",
        "sdkconfig.prod2"
      ],
      "sdkconfigFilePath": "${workspaceFolder}/sdkconfig2"
    },
    "env": {},
    "flashBaudRate": "",
    "idfTarget": "",
    "monitorBaudRate": "",
    "openOCD": {
      "debugLevel": 0,
      "configs": [],
      "args": []
    },
    "tasks": {
      "preBuild": "",
      "preFlash": "",
      "postBuild": "",
      "postFlash": ""
    }
  }
}

```

Note: If using the former ESP-IDF multi config example, remember to remove from root CMakeLists.txt the line:

```
set(SDKCONFIG "${CMAKE_BINARY_DIR}/sdkconfig")
```

[...]

PR #1220, opened 2024-06-17
Fix compiler args for white spaces

radurentea:
## Description

This fixes the bug in which if you open with docker container a project with white spaces in it and try to build the project, it will fail.

Please include a summary of the change and which issue is fixed.

JIRA: [VSC-1381](https://jira.espressif.com:8443/browse/VSC-1381)

## Type of change

Please delete options that are not relevant.
- Bug fix (non-breaking change which fixes an issue)

## Steps to test this pull request

Create project with whitespaces in name "project-n am e" -> re-open in devcontainer -> try to build -> failed.

## How has this been tested?

As described above

**Test Configuration**:
* ESP-IDF Version: 5.2.2
* OS (Windows,Linux and macOS): Windows 11

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test
- [ ] Verified on all platforms - Windows,Linux and macOS

PR #1271, opened 2024-08-05
[VSC-1409] Add check for missing compile_commands.json in IDF projects

radurentea:
## Description

This pull request adds a new feature to improve the user experience when working with ESP-IDF projects in VS Code, particularly when using the Microsoft C/C++ extension.

Key changes:
1. Implements a check for the existence of the `compile_commands.json` file when opening an ESP-IDF project.
2. If the file is missing, displays a notification to the user with an option to generate it.

For users using the Microsoft's C/C++  `compile_commands.json` file is crucial for proper IntelliSense functionality in ESP-IDF projects.

JIRA https://jira.espressif.com:8443/browse/VSC-1409

## Type of change

- New feature (breaking change which adds functionality)

1. Open an ESP-IDF project that doesn't have a `compile_commands.json` file in its `build` directory.
2. Verify that a notification appears, informing about the missing file and offering to generate it.
3. Click the "Generate compile_commands.json" button and confirm that it triggers the `espIdf.idfReconfigureTask` command that generates the `compile_command.json` file in `build` folder of the project.

We should also test the following scenarios:

1. https://github.com/espressif/vscode-esp-idf-extension/pull/1252
2. https://github.com/espressif/vscode-esp-idf-extension/pull/1220
3. Building should be tested for esp-idf v4.3.6 as well (it was reported in the past by a customer that -S was not accepted, only -S=)

## How has this been tested?

As described above

**Test Configuration**:
* ESP-IDF Version: 5.3

## Checklist
- [ ] PR Self Reviewed
- [x] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/src/build/buildTask.ts
+++ b/src/build/buildTask.ts
@@ -23,7 +23,6 @@ import * as vscode from "vscode";
 import * as idfConf from "../idfConfiguration";
 import {
   appendIdfAndToolsToPath,
-  compareVersion,
   getEspIdfFromCMake,
   getSDKConfigFilePath,
   isBinInPath,
@@ -119,17 +118,10 @@ export class BuildTask {
     if (!cmakeCacheExists) {
       const espIdfVersion = await getEspIdfFromCMake(this.idfPathDir);
       let defaultCompilerArgs;
-      const useEqualSign = compareVersion(espIdfVersion, "4.4") >= 0;
       if (espIdfVersion === "x.x") {
         Logger.warn(
           "Could not determine ESP-IDF version. Using default compiler arguments for the latest known version."
         );
-        defaultCompilerArgs = [
-          "-G=Ninja",
-          "-DPYTHON_DEPS_CHECKED=1",
-          "-DESP_PLATFORM=1",
-        ];
-      } else if (useEqualSign) {
         defaultCompilerArgs = [
           "-G=Ninja",
           "-DPYTHON_DEPS_CHECKED=1",
@@ -153,19 +145,11 @@ export class BuildTask {
       }
       let buildPathArgsIndex = compilerArgs.indexOf("-B");
       if (buildPathArgsIndex !== -1) {
-        compilerArgs.splice(buildPathArgsIndex, useEqualSign ? 1 : 2);
+        compilerArgs.splice(buildPathArgsIndex, 2);
       }
-      if (useEqualSign) {
-        compilerArgs.push(`-B=${this.buildDirPath}`);
-      } else {
-        compilerArgs.push("-B", this.buildDirPath);
-      }
+      compilerArgs.push("-B", this.buildDirPath);
       if (compilerArgs.indexOf("-S") === -1) {
-        if (useEqualSign) {
-          compilerArgs.push(`-S=${this.currentWorkspace.fsPath}`);
-        } else {
-          compilerArgs.push("-S", this.currentWorkspace.fsPath);
-        }
+        compilerArgs.push("-S", this.currentWorkspace.fsPath);
       }
 
       const sdkconfigFile = await getSDKConfigFilePath(this.currentWorkspace);
@@ -261,7 +245,9 @@ export class BuildTask {
       this.currentWorkspace
     ) as string;
 
-    const adapterTargetName = await getIdfTargetFromSdkconfig(this.currentWorkspace);
+    const adapterTargetName = await getIdfTargetFromSdkconfig(
+      this.currentWorkspace
+    );
     const showTaskOutput =
       notificationMode === idfConf.NotificationMode.All ||
       notificationMode === idfConf.NotificationMode.Output
--- a/src/build/buildTask.ts
+++ b/src/build/buildTask.ts
@@ -123,12 +123,6 @@ export class BuildTask {
           "Could not determine ESP-IDF version. Using default compiler arguments for the latest known version."
         );
         defaultCompilerArgs = [
-          "-G=Ninja",
-          "-DPYTHON_DEPS_CHECKED=1",
-          "-DESP_PLATFORM=1",
-        ];
-      } else {
-        defaultCompilerArgs = [
           "-G",
           "Ninja",
           "-DPYTHON_DEPS_CHECKED=1",
--- a/src/espIdf/reconfigure/task.ts
+++ b/src/espIdf/reconfigure/task.ts
@@ -27,7 +27,7 @@ import {
   workspace,
 } from "vscode";
 import { NotificationMode, readParameter } from "../../idfConfiguration";
-import { appendIdfAndToolsToPath, compareVersion, getEspIdfFromCMake, getSDKConfigFilePath } from "../../utils";
+import { appendIdfAndToolsToPath, getSDKConfigFilePath } from "../../utils";
 import { join } from "path";
 import { TaskManager } from "../../taskManager";
 import { getVirtualEnvPythonPath } from "../../pythonManager";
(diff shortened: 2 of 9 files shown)


src/logger/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+    await Promise.resolve(action());
+  }
+}
+
+// Helper functions for common actions
+export function executeCommand(
+  commandId: string,
+  ...args: any[]
+): NotificationAction {
+  return () => vscode.commands.executeCommand(commandId, ...args);
```
brianignacio5, line 32: I feel this function is unnecessary just doing `() => vscode.commands.executeCommand(commandId, ...args)` as argument or do `as NotificationAction` in case linter complains.

src/extension.ts
```diff
@@ -3558,6 +3559,44 @@ export async function activate(context: vscode.ExtensionContext) {
       new HintHoverProvider(treeDataProvider)
     )
   );
+
+  checkAndNotifyMissingCompileCommands();
+  context.subscriptions.push(
+    vscode.workspace.onDidChangeWorkspaceFolders(() => {
+      checkAndNotifyMissingCompileCommands();
+    })
```
brianignacio5, line 3567: Why not add this function call in the already existing `vscode.workspace.onDidChangeWorkspaceFolders ` ?

We probably need to wrap that one to `context.subscriptions.push(...)`

src/extension.ts
```diff
@@ -3558,6 +3559,44 @@ export async function activate(context: vscode.ExtensionContext) {
...
+            "build",
+            "compile_commands.json"
+          );
+          const compileCommandsExists = await pathExists(compileCommandsPath);
+
+          if (!compileCommandsExists) {
+            showInfoNotificationWithAction(
+              "compile_commands.json is missing. This may cause errors with the Microsoft C/C++ extension.",
+              "Generate compile_commands.json",
+              executeCommand("espIdf.idfReconfigureTask")
```
brianignacio5, line 3588: Could just rewrite this to `() => vscode.commands.executeCommand("espIdf.idfReconfigureTask")`

src/extension.ts
```diff
@@ -3558,6 +3559,44 @@ export async function activate(context: vscode.ExtensionContext) {
...
+          const compileCommandsPath = path.join(
+            folder.uri.fsPath,
+            "build",
+            "compile_commands.json"
+          );
+          const compileCommandsExists = await pathExists(compileCommandsPath);
+
+          if (!compileCommandsExists) {
+            showInfoNotificationWithAction(
+              "compile_commands.json is missing. This may cause errors with the Microsoft C/C++ extension.",
```
brianignacio5, line 3586: This could be localized with `vscode.l10n.t("text")` using bundles. 

I would also suggest to remove the reference to Microsoft C/C++ extension since compile_commands.json could affect other extensions too such as Clang extension.

brianignacio5 (changes requested):
Few changes I consider would improve this PR

brianignacio5 (approved):
LGTM

radurentea:
Hi @Fabricio-ESP, PTAL

Fabricio-ESP:
The extension is properly detecting the missing file in the builds folder and running the reconfigure command. Tested on Windows and Linux.

Side effect is that when opening a new project (from a example) the build folder does not exist yet, and the error notification will be triggered. Running the reconfigure command will generate the build for the ESP32 default target. If the user wants to change the target the build folder will need to recreated.

src/extension.ts
```diff
@@ -3568,6 +3575,40 @@ export async function activate(context: vscode.ExtensionContext) {
...
+  checkAndNotifyMissingCompileCommands();
+}
+
+function checkAndNotifyMissingCompileCommands() {
+  if (vscode.workspace.workspaceFolders) {
+    vscode.workspace.workspaceFolders.forEach(async (folder) => {
+      try {
+        const isIdfProject = utils.checkIsProjectCmakeLists(folder.uri.fsPath);
+        if (isIdfProject) {
+          const compileCommandsPath = path.join(
```
brianignacio5, line 3587: I will suggest using the build path setting instead of static build directory.

```ts
const buildDirPath = idfConf.readParameter(
        "idf.buildPath",
        workspaceRoot
      ) as string;
```

brianignacio5 (changes requested):

Fabricio-ESP (approved):
Tested on Windows, WSL and Linux. 
Reconfigure command working as expected.

src/extension.ts
```diff
@@ -391,77 +392,81 @@ export async function activate(context: vscode.ExtensionContext) {
     binTimestampEventFunc
   );
   context.subscriptions.push(buildWatcherCreateDisposable);
-
```
brianignacio5, line 394: Changes from 394 to 469 are not related to the PR proposal solution. While I think is good to add this to `context subscription` disposable array, the amount of changes are not necessary. 

Please update to remove these changes.

brianignacio5 (changes requested):
Rebase issues.

src/build/buildTask.ts
```diff
@@ -129,12 +127,6 @@ export class BuildTask {
           "-DPYTHON_DEPS_CHECKED=1",
           "-DESP_PLATFORM=1",
         ];
-      } else if (useEqualSign) {
```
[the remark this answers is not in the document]
radurentea: These changes were intentional and are part of the commit [Remove unnecessary validation - c8f9ec8](https://github.com/espressif/vscode-esp-idf-extension/pull/1271/commits/c8f9ec8efe1ba98b92c84ffee8b00fa0de43ec65)

After some testing, I found out the implementation with "=" was unnecessary. After the changes we've made with the configuration settings. These commands are being run in the python env, so the behaviour is the same for all shells the "=" was added for some specific case Andrii found in the past with consecutive white spaces in the name.

src/espIdf/reconfigure/task.ts
```diff
@@ -65,18 +65,12 @@ export class IdfReconfigureTask {
...
 
     let buildPathArgsIndex = reconfigureArgs.indexOf("-B");
     if (buildPathArgsIndex !== -1) {
-      reconfigureArgs.splice(buildPathArgsIndex, useEqualSign ? 1 : 2);
-    }
-    if (useEqualSign) {
-      reconfigureArgs.push(`-B=${this.buildDirPath}`);
-    } else {
-      reconfigureArgs.push("-B", this.buildDirPath);
+      reconfigureArgs.splice(buildPathArgsIndex, 2);
```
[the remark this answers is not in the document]
radurentea: Same as above

src/extension.ts
radurentea: Yes, this was the issue I had with the rebase. I think I might have change this part of the code in the past with another rebase. Here is where I've tried to make manual fixes on the past rebase.

I will try to back on past rebases on my local to see if I can fix this the proper way

radurentea: I finally figured it out, the problem is with this commit: https://github.com/espressif/vscode-esp-idf-extension/commit/5de32524f8abfabd59b594aca33b51b88c2716f1

I will revert one commit before and will try to cherry picks the other commits or add the changes from them manually

radurentea: Hi @brianignacio5,

I've done what I've mentioned above, please take another look

src/build/buildTask.ts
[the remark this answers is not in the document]
brianignacio5: I can't see the commit since it is lost maybe due to squash commits.

`These commands are being run in the python env` ? How ? where?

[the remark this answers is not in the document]
brianignacio5: I understand your logic but I think is better in the future to separate this into 2 PRs: 1) compile_commands.json work and 2) The removal of `=`. When I see the changes I just thought of it as a regression because it changes code unrelated to the PR.

We can proceed testing it in this PR but then is better to add how is fixed in the PR description and steps to test. 
I think the issue was an old CMake version (Please test with esp-idf v4.3.6) didn't support -S <path> only -S=<path> and it was reported by a customer, hence my remark to make sure this does work.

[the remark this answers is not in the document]
radurentea: Hi @brianignacio5,

I agree with your vision of splitting PRs, but in this particular case there is one thing that I forgot to mention: which is that I've removed the "=" because there was an issue on this PR that Fabricio found and it was caused by the compiler argument "-B=". 

`Executing task: /home/fabricio/.espressif/python_env/idf5.3_py3.12_env/bin/python /home/fabricio/esp/v5.3.1/esp-idf/tools/idf.py -B=/home/fabricio/Workspace/WSL-Local/blink/build -DSDKCONFIG=/home/fabricio/Workspace/WSL-Local/blink/sdkconfig reconfigure`

image

I will be updating the testing steps in the description.

brianignacio5:
Test building blink on Windows with blink using v5.3.1 and esp-idf 4.3.6 and build works fine.

brianignacio5 (approved):
LGTM

Fabricio-ESP:
@radurentea 
Tested on Windows and Ubuntu, also using docker container and multi configuration.
All working as expected

Fabricio-ESP (approved):

PR #1286, opened 2024-08-22
Fix display of monitor the device when flash fails

radurentea:
## Description

- Changed the following message `You can monitor with ESP-IDF: Monitor your device command` to:
 `You can monitor your device with ESP-IDF: Monitor command`
- Change behaviour so that we don't display the message if flash fails

JIRA: https://jira.espressif.com:8443/browse/VSC-1431

## Type of change

### Preparation:
1. Ensure you have the latest version of the VS Code Extension source code:

    - Install [Node.js](https://nodejs.org/en/)
    - Clone this repository git clone --recursive https://github.com/espressif/vscode-esp-idf-extension.git
    - Install all the dependencies using yarn. (you might need to install yarn)
    - Open the project in Visual Studio Code and press F5 to run with the debugger. This will launch a new VS Code Extension Development Host window, allowing you to debug the extension. (Note: This method allows you to test changes directly, as an alternative to installing the extension via a .vsix file.)
2. Set up a test ESP project that can be flashed.

### Test Scenarios:

A. Normal Behavior (Baseline):
1. Attempt to flash a project that should succeed.
2. Verify that the flashing process completes successfully.
3. Confirm the message "You can monitor your device with ESP-IDF: Monitor command" appears.

B. Simulated Flash Failure:
1. Modify the file `src\flash\uartFlash.ts` in the VS Code Extension source code.
2. Add the following code beneath the `try{` inside `async function flashCommand`:
   ```typescript
   throw new Error("TEST_UNUSUAL_ERROR");

Like this:

```javascript
try {
    throw new Error("TEST_UNUSUAL_ERROR"); // <-- Add this line
    const model = await createFlashModel(
      flasherArgsJsonPath,
      port,
      flashBaudRate
    );
```

## How has this been tested?

As described above

**Test Configuration**:
* ESP-IDF Version: 5.1.4

## Checklist
- [ ] PR Self Reviewed
- [x] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -3920,10 +3920,11 @@ const flash = (
             workspaceRoot
           ) as ESP.FlashType;
         }
-        await startFlashing(cancelToken, flashType, encryptPartition);
-        OutputChannel.appendLine(
-          "Flash has finished. You can monitor with ESP-IDF: Monitor your device command"
-        );
+        if (await startFlashing(cancelToken, flashType, encryptPartition)) {
+          OutputChannel.appendLine(
+            "Flash has finished. You can monitor your device with ESP-IDF: Monitor command"
+          );
+        }
       }
     );
   });
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -3922,7 +3922,7 @@ const flash = (
         }
         if (await startFlashing(cancelToken, flashType, encryptPartition)) {
           OutputChannel.appendLine(
-            "Flash has finished. You can monitor your device with ESP-IDF: Monitor command"
+            "Flash has finished. You can monitor your device with 'ESP-IDF: Monitor command'"
           );
         }
       }


radurentea:
Hi @brianignacio5, PTAL

brianignacio5 (approved):
LGTM

radurentea:
HI @AndriiFilippov, PTAL

src/extension.ts
```diff
@@ -3920,10 +3920,11 @@ const flash = (
             workspaceRoot
           ) as ESP.FlashType;
         }
-        await startFlashing(cancelToken, flashType, encryptPartition);
-        OutputChannel.appendLine(
-          "Flash has finished. You can monitor with ESP-IDF: Monitor your device command"
-        );
+        if (await startFlashing(cancelToken, flashType, encryptPartition)) {
+          OutputChannel.appendLine(
+            "Flash has finished. You can monitor your device with ESP-IDF: Monitor command"
```
AndriiFilippov, line 3925: Shouldn't we highlight the command with ' ' ? 
` "Flash has finished. You can monitor your device with 'ESP-IDF: Monitor' command"`

AndriiFilippov:
@radurentea hi !

Tested under: 
OS - Windows 11 / MacOS 

do see "success" message: 
image

do see "fail" message": 
image

LGTM 👍 


Only question related to output: 
Why don't we clean the console ? 
I mean if I do successful flash and the next flash will fail -  I will get something like this. Where you could see the previous message about successful flash and then the  notification about failed flash 
image

Nothing critical but should we somehow separate this messages from each other with space ? or simply clean console

src/extension.ts
radurentea: Thanks, I think it's a good idea. I've just modified it

PR #1330, opened 2024-10-23
[VSC-1472/1207] add new idf size format support

brianignacio5:
## Description

Support new esp-idf-size JSON format which provide more map file information in `ESP-IDF: Size analysis of the binaries` command UI.

This feature is supported for ESP-IDF v5.3 and higher, when older versions are used the extension will use previous IDF Size UI
Screenshot 2024-10-23 at 17 10 38

## Type of change

1. Click on menu **View**, **Command Palette** and search for `ESP-IDF: Size analysis of the binaries` command with ESP-IDF v5.3.0 or higher
2. Execute action. The rendered UI should show more fields and information.
3. Observe results.
4. Repeat step 1 with ESP-IDF version older than v5.3.0. Results should show previous behavior.

- Expected behaviour:
Result similar to https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/tools/idf-size.html?highlight=size

- Expected output:

https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/tools/idf-size.html?highlight=size

## How has this been tested?

Manual testing of extension command using both v5.4.0 and v.5.2.2

**Test Configuration**:
* ESP-IDF Version: v5.4.0

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/src/espIdf/size/idfSize.ts
+++ b/src/espIdf/size/idfSize.ts
@@ -60,8 +60,10 @@ export class IDFSize {
       ) as string;
       const version = await utils.getEspIdfFromCMake(espIdfPath);
       const formatArgs =
-        utils.compareVersion(version, "5.1.0") >= 0
-          ? ["--format", "json"]
+        utils.compareVersion(version, "5.3.0") >= 0
+          ? ["--format", "json2"]
+          : utils.compareVersion(version, "5.1.0") >= 0
+          ? ["--format", "json"] 
           : ["--json"];
       const overview = await this.idfCommandInvoker([
         "idf_size.py",
--- a/src/espIdf/size/idfSizePanel.ts
+++ b/src/espIdf/size/idfSizePanel.ts
@@ -2,20 +2,19 @@
  * Project: ESP-IDF VSCode Extension
  * File Created: Thursday, 20th June 2019 10:39:58 am
  * Copyright 2019 Espressif Systems (Shanghai) CO LTD
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *    http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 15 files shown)


Fabricio-ESP:
Having trouble to run the command using IDF 5.3.1 on a Windows + WSL environment.
IDF.py size command is successful in the terminal. After the project is built I can see the table printed with some size allocation and an error (straight from IDF) which is probably unrelated. Still the extension shows a pop up message "Error encountered while calling idf_size.py.

image

Running on Windows only I see different contents between IDF 5.2.3:
image

image

Compared to 5.3.1:
image

image

But it seems the submenus are not being displayed for all files, not sure why.

Fabricio-ESP:
The issue seems to be associated to the selected target. For ESP32 > Wrover Kit I see the table being rendered incorrectly on windows, but complete on WSL, while when selecting ESP32C6 I see the error popup without any size table.

radurentea:
1. I've run into the same issue, but it seems this happens only with an example I've already had in my PC. When I have this error, the Size Analysis Webview doesn't open.
image
After reconfiguring the extension with esp-idf 5.3.1 and building a new blink example, I did not have this issue anymore.

2. I've also noticed that for the new Size Analysis, there is only "libesp_app_format.a" which can be extended. Probably there is less data or the data has changed and we don't create the fields properly?

**UI Improvement suggestion for future, not related to this PR**
1. The rows to have 2 different background colors, one darker and one lighter this would help with readability. I think the term is zebra striping.
image

2. The columns to be aligned, when there is a sub-menu like in "libesp_app_format.a" the columns become unaligned with the others
image

brianignacio5:
Share the idf_size error from extension log or doctor command output ?

Fabricio-ESP:
Data is rendering as expected now.

Th eone item that can be treated in other PR, related to customer experience improvement. The size chart and table are not updated if the user re-runs the command before closing the previous results. Either we close it and open a new window with the resilts, or update the results in place. It is not clear to the user that the results presented are outdated.

Fabricio-ESP (approved):

radurentea (approved):
LGTM

PR #1338, opened 2024-11-08
[VSC-1515] fix unit tests not being refresh

### LINKED ISSUE #1336: [Feature Request]: Add button to refresh tests in testing tab (VSC-1515)
The testing tab does not automatically recognize newly created unity tests and requires an IDE restart in order to run new tests through the tab, including test name changes. It would be great if the extension could recognize that a new test was created automatically or if the 'Refresh Tests' button in the testing tab searched through the codebase for new tests. I have asked if there is a way to refresh the tab without an editor restart in the [espressif forums](https://www.esp32.com/viewtopic.php?f=40&t=42590), with no response, so I am assuming this feature doesn't exist. I think that the best way to implement the feature is to use the existing 'Refresh Tests' button for this purpose, unless that is impossible.

brianignacio5:
## Description

Fix unit tests not being refreshed when Refresh tests button is clicked.

Fixes #1336

## Type of change

1. Open a ESP-IDF project like **unity-test** that contains test cases as described in https://docs.espressif.com/projects/vscode-esp-idf-extension/en/latest/additionalfeatures/unit-testing.html
2. Add a new test case in any file test_*.c file
3. Press menu **View** then **Testing** Click the Refresh Tests button in the top of the side (Looks like a loading button)
4. The new test should appear.

- Expected behaviour:
Test cases should be added, deleted or modified when Refresh Tests button is executed and test cases have changed.

- Expected output:

## How has this been tested?

Manual testing using the **unity-test** esp-idf example.

**Test Configuration**:
* ESP-IDF Version: 5.0

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/src/espIdf/unitTest/adapter.ts
+++ b/src/espIdf/unitTest/adapter.ts
@@ -53,6 +53,7 @@ export class UnitTest {
     this.unitTestController.refreshHandler = async (
       cancelToken?: CancellationToken
     ) => {
+      this.clearExistingTestCaseItems();
       const fileList = await getFileList();
       this.testComponents = await getTestComponents(fileList);
       await this.loadTests(fileList);
@@ -160,6 +161,12 @@ export class UnitTest {
     context.subscriptions.push(this.unitTestController);
   }
 
+  clearExistingTestCaseItems() {
+    this.unitTestController.items.forEach((item) =>
+      this.unitTestController.items.delete(item.id)
+    );
+  }
+
   async createFileTestCaseItems(file: Uri) {
     const existing = this.unitTestController.items.get(file.toString());
     if (existing) {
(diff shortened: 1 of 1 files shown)


radurentea:
LGTM

radurentea (approved):

Fabricio-ESP:
Hi @brianignacio5 ,
Confirmed that the new fix capture new test_ files created for the components. I may not be fully familiar with the functionality but is it correct to only look for test_ files in the components folder?

Also noticed that if I rename a test file and try to update the testing tab I see a new entry with the new file name, fur the application does not remove the old entry, so it ends up showing 2 files with the old and new name. If I restart VSCode than it removes the old entry. Is there a reason we would like to keep the old filename on the testing tab?

brianignacio5:
> Hi @brianignacio5 , Confirmed that the new fix capture new test_ files created for the components. I may not be fully familiar with the functionality but is it correct to only look for test_ files in the components folder?

Yes that is correct that is looking for test_* files.

> Also noticed that if I rename a test file and try to update the testing tab I see a new entry with the new file name, fur the application does not remove the old entry, so it ends up showing 2 files with the old and new name. If I restart VSCode than it removes the old entry. Is there a reason we would like to keep the old filename on the testing tab?

Updated the PR to delete all items before creating them on refresh. PTAL again @Fabricio-ESP

Fabricio-ESP (approved):
Confirmed working as expected.
Testing tab was able to locate test files and update the list upon clicking the refresh button.
Note that test files should start with test_ and should be inside a folder called test.
This is not properly set int he unit_test example provided by IDF and might confuse the users.

PR #1343, opened 2024-11-18
[VSC-1521] use idf tools export to get env vars

brianignacio5:
## Description

Use `idf_tools.py export --format key-value` to get ESP-IDF variables required in environment variables.

This should fix missing `ESP_IDF_VERSION` from extension tasks and use PATH from idf_tools.py (which is already done but could replace extension implementation.)

## Type of change

Probably the best way to verify that variables are present is to open `ESP-IDF: Open ESP-IDF Terminal` and do `echo $ESP_IDF_VERSION` or other environment variable from `idf_tools.py export` output.  

You can also test with any command like `ESP-IDF: Build your project` but harder to see if variables are available.

- Expected behaviour:

All variables from `idf_tools.py export --format key-value` should be available in extension commands like `ESP-IDF: Open ESP-IDF Terminal`

**Test Configuration**:
* ESP-IDF Version: 4.4 , 4.4.8, 5.0.4 5.4
* OS (Windows,Linux and macOS): macOS Windows

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/src/pythonManager.ts
+++ b/src/pythonManager.ts
@@ -26,6 +26,7 @@ import { delimiter, dirname, join, sep } from "path";
 import { OutputChannel } from "./logger/outputChannel";
 import { readParameter, writeParameter } from "./idfConfiguration";
 import { ESP } from "./config";
+import { EOL } from "os";
 
 export async function installEspIdfToolFromIdf(
   espDir: string,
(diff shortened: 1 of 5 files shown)


Fabricio-ESP:
I installed the extension and I keep getting error when attempting to build the example project.
image
[Error after build.txt](https://github.com/user-attachments/files/17817534/Error.after.build.txt)

brianignacio5:
What is the build output ?

radurentea (approved):
LGTM

I would have liked if the ESP-IDF version would return the entire version, but I see this is what is being returned by the "export" command. I've tested it with  esp idf  "5.3.1" version and the version I see in doctor command is "5.3"

Fabricio-ESP (approved):
Confirmed working on Windows and Linux, running IDF 5.0, 5.3 and 5.4

PR #1360, opened 2024-12-04
[VSC-1534] Fix Install Extension Python Requirements

radurentea:
## Description

Bug: Constraints File Path Mismatch

The extension currently searches for ESP-IDF constraint files using the full version number (including patch version), but ESP-IDF only uses major.minor versions in the filename.

Example:
- Current code looks for: `espidf.constraints.v5.3.1`
- Actual ESP-IDF filename: `espidf.constraints.v5.3`

This PR fixes the version parsing to correctly match ESP-IDF's constraint file naming convention.

This PR changes this behavior.
## Type of change

1) Have a project created with v5.3.1 of esp-idf.
2) Build + select device target
3) Use the command from ESP-IDF Explorer: Advanced > Install Extension Python Requirements
4) Try to change device target (it should work)

Before this bugfix, changing the device target would not work because of the error in the logs: `The following Python requirements are not satisfied`

## How has this been tested?

As described above

**Test Configuration**:
* ESP-IDF Version: 5.3.1

## Checklist
- [ ] PR Self Reviewed
- [x] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/src/pythonManager.ts
+++ b/src/pythonManager.ts
@@ -182,7 +182,12 @@ export async function installExtensionPyReqs(
     OutputChannel.appendLine(debugAdapterRequirements + reqDoesNotExists);
     return;
   }
-  const espIdfVersion = await utils.getEspIdfFromCMake(espDir);
+  const fullEspIdfVersion = await utils.getEspIdfFromCMake(espDir);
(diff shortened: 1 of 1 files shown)


brianignacio5 (approved):
LGTM

PR #1385, opened 2024-12-26
[VSC-1537] use IDF_TARGET in customExtraVars if no sdkconfig

### LINKED ISSUE #1363: Can't switch target on v1.9.0 (VSC-1537)
### OS

MacOS

### Operating System version

15.1.1 (24B2091)

1.9.0

### ESP-IDF version

5.3.1

### Python version

3.8.7

### Doctor command output

NA

### Extension

_No response_

### Description

Project will not switch back to esp32s3 after going to esp32 target. Downgraded to 1.8.1 and it worked. 

### Debug Message

NA

### Other Steps to Reproduce

_No response_

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

### LINKED ISSUE #1386: [Feature Request]: Allow 'idf.svdFilePath' to construct the path to an .svd file based on the current target (VSC-1550)
**Is your feature request related to a problem?**

The "ESP-IDF Peripheral Viewer" window requires the setting `idf.svdFilePath` to be set to the full path of an SVD file matching the target being debugged (i.e. esp32s3 or esp32 ...). Before v1.9.0, `idf.adapterTargetName` could be used to construct `idf.svdFilePath` as follows:
```json
"idf.svdFilePath": "${workspaceFolder}/svd/${config:idf.adapterTargetName}.svd"
```
This made configuring the svd file path automatic even when the target changed. There was never any need to manually adjust `idf.svdFilePath`. 

Version v1.9.0, removed `idf.adapterTargetName` (a *great change* in my opinion). This makes it impossible to set `idf.svdFilePath` to automatically adapt to the target.

This feature proposes a few ways to bring back automatic configuration of `idf.svdFilePath` as the target changes without the problematic setting `idf.adapterTargetName`.

Expose a read only configuration setting tentatively called `idf.currentTargetName` which is interpreted as the value returned by `getIdfTargetFromSdkconfig()`. This would allow the value of `idf.svdFilePath` to automatically adapt as the target changes with the following sample configuration in `settings.json`:
```json
"idf.svdFilePath": "${workspaceFolder}/svd/${config:idf.currentTargetName}.svd"
```

**Describe alternatives you've considered**

* Another approach would be to add a new setting tentatively called `idf.svdFilesRoot` pointing to a directory where the extension can locate the svd file, assuming it is called getIdfTargetFromSdkconfig() + '.svd'.

* Another approach would be to add a command tentatively called `idf.getCurrentTargetName` which would return whatever `getIdfTargetFromSdkconfig()` returns at the time `idf.svdFilePath` is being used by the extension. The following sample configuration in `settings.json` would be valid:
    ```json
    "idf.svdFilePath": "${workspaceFolder}/svd/${command:idf.getCurrentTargetName}.svd"
    ```
* Another option would be to allow accessing SDK configuration values like `IDF_TARGET` from settings.json, with a syntax similar to `${sdkConfig:IDF_TARGET}`. The following sample configuration in `settings.json` would be valid:
    ```json
    "idf.svdFilePath": "${workspaceFolder}/svd/${sdkConfig:IDF_TARGET}.svd"
    ```

All approaches maintain the ability to set `idf.svdFilePath` to a fixed, non-target dependent path for users who prefer this option.

brianignacio5:
## Description

To find the current IDF_TARGET the extension reads the sdkconfig in v1.9.0 but before there was a `idf.adapterTargetName` configuration setting which was used to resolve IDF_TARGET. The previous logic would fail if the current project is configure for a target (say esp32) but the configuration setting is set to (say esp32c3).

Reading from sdkconfig seems like a better approach, but what happens when you want to define a IDF_TARGET but no sdkconfig exists ? For this case we still need to persist this somehow. Now we would use `idf.customExtraVars` configuration setting and add `IDF_TARGET` when running the `ESP-IDF: Set Espressif Device Target` command. The extension will use this value when sdkconfig IDF_TARGET value doesn't exist.

Also add `idf.customExtraVars` to idfConfiguration.ts readParameter. When a user has a setting with value ${env:VARNAME} now it will also use `idf.customExtraVars` to resolve it. For example `"idf.svdFilePath": "${workspaceFolder}/${env:IDF_TARGET}.svd"`

Fixes #1363 
Fixes #1386

## Type of change

1. Run the `ESP-IDF: Set Espressif Device Target` on a ESP-IDF project. This will create an sdkconfig file and build project.
2. Delete the build directory and sdkconfig file.
3. Run the `ESP-IDF: Build your project` command. The project should build with the target defined before instead of previously default esp32.

- Expected behaviour:

The IDF_TARGET that was set in 1 is still used when sdkconfig has been deleted for other extension commands like `ESP-IDF: Build your project`.

- Expected output:

The IDF_TARGET that was set in 1 is still used when sdkconfig has been deleted for other extension commands like `ESP-IDF: Build your project`.

## How has this been tested?

Manual test by following steps above.

**Test Configuration**:
* ESP-IDF Version: 5.3.1

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting
- [x] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/README.md
+++ b/README.md
@@ -555,8 +555,9 @@ We have implemented some utilities commands that can be used in `tasks.json` and
 - `espIdf.getOpenOcdScriptValue`: Return the value of OPENOCD_SCRIPTS computed from ESP-IDF Tools path, `idf.customExtraVars`, or the system's OPENOCD_SCRIPTS environment variable.
 - `espIdf.getOpenOcdConfig`: Return the openOCD configuration files as string. Example `-f interface/ftdi/esp32_devkitj_v1.cfg -f board/esp32-wrover.cfg`.
 - `espIdf.getProjectName`: Return the project name from current workspace folder `build/project_description.json`.
-- `espIdf.getToolchainGcc`: Return the absolute path of the toolchain GCC for the ESP-IDF target given by current IDF_TARGET in sdkconfig.
-- `espIdf.getToolchainGdb`: Return the absolute path of the toolchain gdb for the ESP-IDF target given by current IDF_TARGET in sdkconfig.
+- `espIdf.getToolchainGcc`: Return the absolute path of the toolchain GCC for the ESP-IDF target given by current IDF_TARGET in sdkconfig or `idf.customExtraVars`["IDF_TARGET"] configuration setting.
+- `espIdf.getToolchainGdb`: Return the absolute path of the toolchain gdb for the ESP-IDF target given by current IDF_TARGET in sdkconfig or `idf.customExtraVars`["IDF_TARGET"] configuration setting.
+- `espIdf.getIDFTarget`: Return the current IDF_TARGET from sdkconfig or `idf.customExtraVars`["IDF_TARGET"] configuration setting.
 
 See an example in the [debugging](https://docs.espressif.com/projects/vscode-esp-idf-extension/en/latest/debugproject.html) documentation.
 
--- a/README_CN.md
+++ b/README_CN.md
@@ -555,8 +555,9 @@ ESP-IDF 扩展在 VS Code 底部蓝色窗口的状态栏中提供了一系列命
 - `espIdf.getOpenOcdScriptValue`：返回从 ESP-IDF 工具路径、`idf.customExtraVars` 或系统 OPENOCD_SCRIPTS 环境变量中计算出的 OPENOCD_SCRIPTS 的值。
 - `espIdf.getOpenOcdConfig`：以字符串形式返回 openOCD 配置文件。例如 `-f interface/ftdi/esp32_devkitj_v1.cfg -f board/esp32-wrover.cfg`。
 - `espIdf.getProjectName`：从当前工作区文件夹的 `build/project_description.json` 文件中提取项目名称。
-- `espIdf.getToolchainGcc`：根据 sdkconfig 文件中指定的 IDF_TARGET，该命令将返回相应 GCC 工具链的绝对路径。
-- `espIdf.getToolchainGdb`：根据 sdkconfig 文件中指定的 IDF_TARGET，该命令将返回相应 GDB 工具链的绝对路径。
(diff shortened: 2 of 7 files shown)


Fabricio-ESP:
Hi Brian,
What are the other 2 parameters added to the customExtraVars? Part of this solution, or anticipation for other changes?

    "idf.customExtraVars": {
        "OPENOCD_SCRIPTS": "/home/virtual/esp/v5.0.7/Tools/tools/openocd-esp32/v0.12.0-esp32-20240318/openocd-esp32/share/openocd/scripts",
        "ESP_ROM_ELF_DIR": "/home/virtual/esp/v5.0.7/Tools/tools/esp-rom-elfs/20220823/",
        "IDF_TARGET": "esp32c3"
   }     
   
Tested on Windows and Ubuntu, it is working. The scenario with deleted sdkconfig was not considered with the previous change.

Fabricio-ESP (approved):

brianignacio5:
> Hi Brian, What are the other 2 parameters added to the customExtraVars? Part of this solution, or anticipation for other changes?
> 
> ```
> "idf.customExtraVars": {
>     "OPENOCD_SCRIPTS": "/home/virtual/esp/v5.0.7/Tools/tools/openocd-esp32/v0.12.0-esp32-20240318/openocd-esp32/share/openocd/scripts",
>     "ESP_ROM_ELF_DIR": "/home/virtual/esp/v5.0.7/Tools/tools/esp-rom-elfs/20220823/",
>     "IDF_TARGET": "esp32c3"
> ```
> 
> }
> 
> Tested on Windows and Ubuntu, it is working. The scenario with deleted sdkconfig was not considered with the previous change.

This values were being set before this PR when configuring the extension. These env variables are generated from $IDF_PATH/tools/tools.json and also added in env vars in IDF when running $IDF_PATH/export.sh

PR #1476, opened 2025-03-14
Enhance/hints viewer openocd

radurentea:
## Description

This enhancement:
- adds hints for openOCD errors (manual searching using esp-idf hints viewer, search error command)
- improves matching for building errors based on automatic/manual searches
- adds notification system that will show the user 2 actions when it comes to hints related to building the project
      - Focus on the  ESP-IDF hints bottom panel
      - Mutes "Found Hint" notification for this VS Code session (until VS Code is closed)
- adds real-time hints display for openocd errors (when running OpenOCD session, if an error appears and a hint is found for that error, Hints Viewer bottom panel will be focused)
- improves hover over errors in text editor (building errors) by increasing the display of the hints when hovering over the entire line where error was found. In the old version, the squiggly line had the length of 1 character. If there were other errors in the same line, for example from C/C++ Extension, it would require luck to find the portion for the ESP-IDF hint to be shown.
- Adds extra child for hints in the tree structured data which is clickable (will open the link address in default browser), in case the hint has a link property

## Type of change

1. Manual searching for errors related to OpenOCD (OpenOCD version needs to be >= `v0.12.0-esp32-20250226`). List of error-hints pair is taken from:
`<ESP_TOOLS_INSTALLATION_PATH>/tools/openocd-esp32/<OPEN_OCD_VERSION>/openocd-esp32/share/openocd/espressif/tools/esp_problems_hints.yml `
2. Manual searching for errors that come from esp-idf's hints yml file, located at `<ESP_IDF_INSTALLATION_PATH>/<ESP_IDF_VERSION>/esp-idf/tools/idf_py_actions/hints.yml`
3. Try to reproduce errors from inside both yaml files (esp-idf, openocd) and see if the hints are triggered automatically. The errors from esp-idf should display a notification with 2 actions, while the openocd hints should automatically trigger focus on hints viewer bottom panel
4. When having errors inside the text editor, post build (for example: adding #include "spiram.h" inside the main.c file of an esp-idf example project), make sure hint is displayed when hovering over the line that has the squiggly lined text

## How has this been tested?

As described above

**Test Configuration**:
* ESP-IDF Version: master

## Checklist
- [ ] PR Self Reviewed
- [x] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test


.vscode/settings.json
```diff
@@ -35,5 +35,8 @@
   ],
   "typescript.updateImportsOnFileMove.enabled": "always",
   "workbench.editor.enablePreview": true,
-  "iis.configDir": ""
+  "iis.configDir": "",
+  "idf.pythonInstallPath": "/opt/homebrew/bin/python3",
+  "idf.espIdfPath": "/Users/radurentea/esp/v5.4/esp-idf",
+  "idf.toolsPath": "/Users/radurentea/esp/esptools"
```
brianignacio5, line 41: Please remove local settings set here.

src/espIdf/hints/index.ts
```diff
@@ -239,25 +581,99 @@ export class HintHoverProvider implements vscode.HoverProvider {
...
+  try {
+    const idfToolsPath = idfConf.readParameter(
+      "idf.toolsPath",
+      workspace
+    ) as string;
+
+    const openOCDManager = OpenOCDManager.init();
+    const version = await openOCDManager.version();
+
+    if (!version) {
```
brianignacio5, line 650: I think running openOCD just to see if the openOCD hints file exists is not necessary. A simply `pathExists` should be faster and easier no ?

src/espIdf/hints/index.ts
```diff
@@ -239,25 +581,99 @@ export class HintHoverProvider implements vscode.HoverProvider {
...
+      Logger.infoNotify(
+        "Could not determine OpenOCD version. Hints file won't be loaded."
+      );
+      return null;
+    }
+
+    const hintsPath = path.join(
```
brianignacio5, line 657: Also getting the openOCD Hints YAML path is duplicated both here and in the `openocdHInt.ts`. 

We could export the function and reuse it here ?

src/espIdf/hints/openocdhint.ts
```diff
@@ -0,0 +1,332 @@
...
+        workspaceRoot
+      );
+    }
+    return OpenOCDErrorMonitor.instance;
+  }
+
+  public async initialize(): Promise<void> {
+    try {
+        // Check OpenOCD version first
+        const openOCDManager = OpenOCDManager.init();
```
brianignacio5, line 71: I think you can check the yaml file exists first and then check openOCD version. Actually if openOCD has the file, we could argue that openOCD does support the feature.

We could think about this, it could simplify load time here.

brianignacio5 (changes requested):
Some code suggestions to simplify functionality.

src/espIdf/hints/openocdhint.ts
brianignacio5: You are right sorry it does need the version.

src/extension.ts
```diff
@@ -294,6 +302,9 @@ export async function activate(context: vscode.ExtensionContext) {
       vscode.commands.registerCommand(name, telemetryCallback)
     );
   };
+  // Store display hints notification (until VS Code is closed)
+  context.workspaceState.update("idf.showHintsNotification", true);
```
brianignacio5: Why is this state necessary? I don't see being disabled on extension deactivation. Maybe is not necessary ?

brianignacio5 (approved):
LGTM without using the workspaceState setting, which doesn't seem to be configurable.

src/extension.ts
radurentea: The idea behind it was to have a reset of the hints viewer's notifications for each session. I don't remember why I went with this implementation of the Notification System for Hints Viewer.

radurentea: I've pushed a new commit in which I converted idf.showHintsNotification from runtime workspace state to proper configurable setting that will persist in workspace settings

package.json
```diff
@@ -1202,6 +1202,12 @@
             "default": false,
             "description": "%param.useSerialPortVendorProductFilter.title%"
           },
+          "idf.showHintsNotification": {
+            "type": "boolean",
+            "scope": "resource",
+            "default": true,
+            "description": "%param.showHintsNotification%"
```
brianignacio5: Missing translations for param.showHintsNotification

brianignacio5: Also missing adding the setting to settings.rst for documentation.

radurentea: Thanks! I've added the translations in the wrong file

radurentea: Thanks!

brianignacio5 (approved):
LGTM

Fabricio-ESP (approved):
Confirmed working.
Thanks for the change on how the hints are displayed.

src/extension.ts
```diff
@@ -3793,6 +3761,15 @@ export async function activate(context: vscode.ExtensionContext) {
         new HintHoverProvider(treeDataProvider)
       )
     );
+
+    // --- Hints Status Bar Item ---
```
brianignacio5: Please move the hints status bar item creation to createCmdsStatusBarItems in src/statusBar/index.ts.

Please observer we use a dictionary `statusBarItems: { [key: string]: StatusBarItem }` to keep track of enable/disable state in Espressif tab.

src/statusBar/index.ts
```diff
@@ -256,3 +256,50 @@ export function createStatusBarItem(
   }
   return statusBarItem;
 }
+
+// --- ESP-IDF Hints Status Bar Item ---
+let hintsStatusBarItem: StatusBarItem | undefined;
```
brianignacio5: Remove in favor of already existing `statusBarItems: { [key: string]: StatusBarItem }` dictionary.

src/extension.ts
```diff
@@ -3668,16 +3672,19 @@ export async function activate(context: vscode.ExtensionContext) {
     // Register commands for clearing error hints
     vscode.commands.registerCommand("espIdf.errorHints.clearAll", () => {
```
brianignacio5: Please using `registerIDFCommand` instead of `vscode.commands.registerCommand`

src/extension.ts
```diff
@@ -3668,16 +3672,19 @@ export async function activate(context: vscode.ExtensionContext) {
...
       treeDataProvider.clearErrorHints(true); // Clear both build and OpenOCD errors
+      updateHintsStatusBarItem(false);
     });
 
     vscode.commands.registerCommand("espIdf.errorHints.clearBuildErrors", () => {
       treeDataProvider.clearErrorHints(false); // Clear only build errors
+      updateHintsStatusBarItem(false);
     });
 
     vscode.commands.registerCommand(
```
brianignacio5: Same here please using `registerIDFCommand` instead of `vscode.commands.registerCommand`

brianignacio5 (changes requested):
Please review existing status bar items implementation and use of `registerIDFCommand` instead of `vscode.commands.registerCommand`

src/statusBar/index.ts
```diff
@@ -235,6 +236,14 @@ export async function createCmdsStatusBarItems(workspaceFolder: Uri) {
     89,
     commandDictionary[CommandKeys.CustomTask].checkboxState
   );
+  statusBarItems["hints"] = createStatusBarItem(
+    l10n.t("💡 New ESP-IDF Hints!"),
+    l10n.t("ESP-IDF: Hints available. Click to view."),
+    "espIdf.errorHints.focus",
+    1000,
```
brianignacio5, line 243: While I think is fine like this, I would suggest moving this information to `src/cmdTreeView/cmdStore.ts` where you can add `CommandKeys.Hints` and reference it as other commands do it.

The main idea is to have a single place of command description, so translations and command IDs are defined once and re-used in many places. Should make it easier to maintain if we need to update.

brianignacio5 (approved):
LGTM with minor change regarding code organization.

src/statusBar/index.ts
radurentea: I thought of that, but this is not a command per se. The status bar item only appears when hints are available, if there are no hints, the status bar item won't appear.

I don't think it should be displayed as a checkbox inside the Espressif Explorer, but this is just a personal opinion and I can make the change if you still think it would be better to be part of `src/cmdTreeView/cmdStore.ts`

brianignacio5: Ah understood! I agree with you point of view.

PR #1557, opened 2025-05-19
Feature/add devkits support

radurentea:
## Description

>WARNING: This PR should not be merged until a new release from OpenOCD (current version is [v0.12.0-esp32-20250422](https://github.com/espressif/openocd-esp32/releases/tag/v0.12.0-esp32-20250422))

With this new feature, when we set the device target, we will now run a script which will detect all the connected boards. In the background this will run OpenOCD, which will make showing the possible device target list a bit slower.

There are 2 validations happening:

1. We check that the script `<idf_tools_path>/tools/openocd-esp32/<version>/openocd-esp32/share/openocd/espressif/tools/esp_detect_config.py` exists
2. There are no debugging active sessions. (Script doesn't work correctly if debugging session is active)

## Type of change

1. Have an ESP device target connected to the PC.
2. Change the script in `idf_tools_path>/tools/openocd-esp32/<version>/openocd-esp32/share/openocd/espressif/tools/esp_detect_config.py` with the one from [openocd's master branch](https://github.com/espressif/openocd-esp32/blob/master/tools/esp_detect_config.py)
3. In an opened project, press the status bar icon to set the device target
4. Connected devices should appear in the list at the top, having their own separated section. They can also be identified by their "Status: CONNECTED" text.
5. You should be able to build, flash, monitor and debug the selected device without any problems.

Another scenario to test is: if the OpenOCD version is correct but the Python script fails (e.g., not found), the device target list needs to falls back to the previous implementation.

## How has this been tested?

As described above.

**Test Configuration**:
* ESP-IDF Version: 5.5

## Checklist
- [ ] PR Self Reviewed
- [x] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/src/espIdf/setTarget/setTargetInIdf.ts
+++ b/src/espIdf/setTarget/setTargetInIdf.ts
@@ -75,6 +75,7 @@ export async function setTargetInIDF(
 
     setTargetArgs.push("set-target", selectedTarget.target);
     const pythonBinPath = await getVirtualEnvPythonPath(workspaceFolder.uri);
+    OutputChannel.appendLine("Running IDF Set Target action", "Set Target");
     const setTargetResult = await spawn(pythonBinPath, setTargetArgs, {
       cwd: workspaceFolder.uri.fsPath,
       env: modifiedEnv,


brianignacio5 (approved):
LGTM

An small thing, in `setTargetInIDF.ts` we should add a line before spawn (line 78) to print something in OutputChannel  (`OutputChannel.appendLine("Running IDF Set Target action", "Set Target");` for example) to add a separation between set-target output. 

Right now the Detect OpenOCD Config tag is closing after set target output (because spawn also print in OutputChannel).

Issue #1414, opened 2025-01-29
Automatically Detect Project Files to Activate ESP-IDF Plugin


Xinyu-Sheng:
When developing projects based on ESP-IDF using VS Code, it is currently necessary to manually activate the ESP-IDF plugin each time to load the corresponding environment and tools, which is not very convenient. In contrast, PlatformIO can automatically activate its plugin by detecting specific ini files, greatly simplifying the development process. I hope that the ESP-IDF plugin could offer similar functionality, where it would automatically activate upon detecting certain project files (such as sdkconfig or other ESP-IDF-related configuration files) in the current directory, thereby enhancing work efficiency.

I would like to see an implementation where, when a user opens a folder containing an ESP-IDF project, VS Code can automatically recognize the project and activate the ESP-IDF plugin without manual intervention. Specifically, this could be triggered by monitoring the presence of specific files (e.g., sdkconfig, particular content within CMakeLists.txt, etc.) as conditions for activating the plugin.

Currently, one alternative is to manually activate the ESP-IDF plugin through the command palette every time, which is not ideal for developers frequently switching between projects. Another possibility is writing custom scripts to check for these files and activate the plugin, but this adds extra learning and maintenance costs.

Providing such functionality would make the development experience smoother, especially important for those managing multiple different types of projects (including non-ESP-IDF projects) simultaneously. It is hoped that this improvement will be seen in future versions. Additionally, if there are suggestions or discussions on how to better integrate this feature, I am also very willing to participate and contribute.

Xinyu-Sheng:
@radurentea 
I would like to have a stricter detection mechanism because my other projects may currently activate the ESP-IDF extension as they meet the existing rules. To avoid this,could we implement more stringent activation conditions? For example, requiring the presence of multiple specific files unique to ESP-IDF projects, or introducing an additional project-specific identifier that must be present in the workspace for the extension to activate. This would help ensure that the ESP-IDF extension only activates for intended ESP-IDF projects and not for others inadvertently.

Ishan-Sharma-17:
Hello, the change works perfectly now and I no longer see the extension active when opening an STM32 project.

Thank you!

PR #1568, opened 2025-05-23
[VSC-1578] Modify Event Activation for file types

radurentea:
Removed previous activation events for file types:
```
"workspaceContains:**/project_description.json",
"workspaceContains:**/sdkconfig",
"workspaceContains:**/CMakeLists.txt"
```

We only check for CMakeLists.txt at root project folder if it includes `"include($ENV{IDF_PATH}/tools/cmake/project.cmake)"`

## Description

Please include a summary of the change and which issue is fixed.

Fixes [#1414 
](https://github.com/espressif/vscode-esp-idf-extension/issues/1414)
## Type of change

1. When opening an ESP-IDF project, the extension should still activate itself (ex: commands from esp-idf should be visible in status bar)
2. When opening a folder that it's not an esp-idf project there should be two scenarios:
     a. If it contains a `CMakeLists.txt` at the root level which doesn't contain the text `include($ENV{IDF_PATH}/tools/cmake/project.cmake)` extension should ask, using a notification, if you want to activate esp-idf extension anyway.
     b. If it doesn't contain a `CMakeLists.txt` at all, it should not activate automatically.
3. This feature should work with workspaces that include multiple esp-idf projects. If at least one of the projects in the workspace contains a valid esp-idf project, the extension should activate itself.

## How has this been tested?

As described above

**Test Configuration**:
* ESP-IDF Version: 5.4
* OS (Windows,Linux and macOS): Mac

## Checklist
- [ ] PR Self Reviewed
- [x] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -277,31 +277,35 @@ export async function activate(context: vscode.ExtensionContext) {
   utils.setExtensionContext(context);
   ChangelogViewer.showChangeLogAndUpdateVersion(context);
 
-  // Check for CMakeLists.txt and its content before full activation
-  if (PreCheck.isWorkspaceFolderOpen()) {
-    const cmakeListsFiles = await vscode.workspace.findFiles(
-      "**/CMakeLists.txt",
-      "**/{build,node_modules}/**"
-    );
-
+  // Check for root CMakeLists.txt and its content before full activation
+  if (PreCheck.isWorkspaceFolderOpen() && vscode.workspace.workspaceFolders) {
     let hasValidIdfProject = false;
-    for (const cmakeList of cmakeListsFiles) {
-      try {
-        const cmakeContent = await readFile(cmakeList.fsPath, "utf-8");
-        if (
-          cmakeContent.includes(
-            "include($ENV{IDF_PATH}/tools/cmake/project.cmake)"
-          )
-        ) {
-          hasValidIdfProject = true;
-          break; // Found a valid project, activate immediately
+
+    for (const workspaceFolder of vscode.workspace.workspaceFolders) {
+      const rootCMakeListsPath = path.join(
+        workspaceFolder.uri.fsPath,
+        "CMakeLists.txt"
+      );
+      const rootCMakeListsExists = await pathExists(rootCMakeListsPath);
+
+      if (rootCMakeListsExists) {
+        try {
+          const cmakeContent = await readFile(rootCMakeListsPath, "utf-8");
+          if (
+            cmakeContent.includes(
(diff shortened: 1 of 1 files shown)


brianignacio5 (approved):
LGTM

src/extension.ts
```diff
@@ -276,6 +276,57 @@ export async function activate(context: vscode.ExtensionContext) {
   Telemetry.init(idfConf.readParameter("idf.telemetry") || false);
   utils.setExtensionContext(context);
   ChangelogViewer.showChangeLogAndUpdateVersion(context);
+
+  // Check for CMakeLists.txt and its content before full activation
+  if (PreCheck.isWorkspaceFolderOpen()) {
+    const cmakeListsFiles = await vscode.workspace.findFiles(
+      "**/CMakeLists.txt",
+      "**/{build,node_modules}/**"
```
brianignacio5: Why do we need `node_modules`? Isn't it from JavaScript Typescript projects ?

brianignacio5: The build folder is also used in CMake project or many other kind of projects. The next validation is only for CMakeLists.txt so I'm just wondering why other files are needed here.

brianignacio5 (changes requested):
The PR validates CMakeLists.txt only. Using `"**/{build,node_modules}/**"` seems unnecessary.

src/extension.ts
radurentea: The vscode.workspace.findFiles expects as a second argument an exclusion pattern. I just thought it would be good to ignore searching in file such as "node_modules" and "build", if they are present in the opened workspace.

This was the exact thing I wanted to discuss about this new implementation: I can remove the "node_modules" if you think it's unnecessary or add these exclusion patterns for other languages as well, to increase performance of `vscode.workspace.findFiles`

brianignacio5: Oh I see. 

Remember that the extension is activated either by running command or `"workspaceContains:**/CMakeLists.txt"`

Later you can use loop of all workspaceFolders with a relative pattern per workspace folder validation for root only:

```ts
const relativePattern = new RelativePattern(workspaceFolderUri, "CMakeLists.txt");
const files = await workspace.findFiles(relativePattern);
```
The content you are validating is for ESP-IDF root CMakeLists.txt only, so no need to check additional places.

radurentea: I have changed the implementation once again based on our discussion only needing the CMakeLists.txt from root level. There is no need to use workspace.findFiles() anymore. I've left more details in the commit description

PR #1583, opened 2025-06-03
Extend default jtag flash arguments 

erhankur:
## Description

The current JTAG flash process doesn't optimize for transfer speed and efficiency, potentially leading to unnecessary data transfers during flashing.

Added  two new default arguments to the JTAG flash process:
1. `compress`: Reduces the amount of data that needs to be transferred during flashing
2. `skip_loaded`: Verifies flash content first and skips loading sections that already match

### DIFF
--- a/package.json
+++ b/package.json
@@ -1250,6 +1250,8 @@
             "scope": "resource",
             "default": [
               "verify",
+              "compress",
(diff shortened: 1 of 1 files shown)


brianignacio5 (approved):
LGTM

PR #1614, opened 2025-07-11
[VSC-1672] fix: append Git and Pigweed paths to PATH instead of prepending

### LINKED ISSUE #1539: `pathToGitDir` may overwrite the previously set python interpreter path (VSC-1672)
### OS

Linux

### Operating System version

Ubuntu22.04 WSL

1.99.3

### ESP-IDF version

5.4.1

### Python version

3.10

### Doctor command output

None (no need to offer)

### Extension

_No response_

### Description

Currently the `createIdfTerminal` command calls `utils.appendIdfAndToolsToPath` to set additional env variables. However, if `git` and system `python3` are installed under the same directory(in my case, both under `/usr/bin`), the following code will overwrite the previously set python interpreter path for expected venv used by esp-idf:
```javascript
  if (pathToGitDir) {
    modifiedEnv[pathNameInEnv] =
      pathToGitDir + path.delimiter + modifiedEnv[pathNameInEnv];
  }
```

### Debug Message

None

### Other Steps to Reproduce

_No response_

radurentea:
## Description

Previously, the extension prepended the Git path (and Pigweed path, if present) to the PATH environment variable. This could override the Python interpreter from the ESP-IDF virtual environment and other critical tools, especially if the Git path points to a system directory like /usr/bin.

Now, both the Git and Pigweed paths are appended to PATH, ensuring that the ESP-IDF toolchain and Python venv remain the highest priority. This prevents accidental overrides and resolves issues where the wrong Python interpreter is used in the IDF terminal.

Fixes #1539

## Type of change

Open the settings and change the git path to "/usr/bin/git"
image (1)

The expected behavior is that the `/usr/bin` should not be before ESP-IDF toolchain or Python venv in the $PATH.

## How has this been tested?

As described above

**Test Configuration**:
* ESP-IDF Version: 6.0.0
* OS (Windows,Linux and macOS): WSL

## Checklist
- [ ] PR Self Reviewed
- [x] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1199,10 +1199,16 @@ export async function appendIdfAndToolsToPath(curWorkspace: vscode.Uri) {
     }
   }
 
-  if (pathToGitDir && !modifiedEnv[pathNameInEnv].split(path.delimiter).includes(pathToGitDir)) {
+  if (
+    pathToGitDir &&
+    !modifiedEnv[pathNameInEnv].split(path.delimiter).includes(pathToGitDir)
+  ) {
     modifiedEnv[pathNameInEnv] += path.delimiter + pathToGitDir;
   }
-  if (pathToPigweed && !modifiedEnv[pathNameInEnv].split(path.delimiter).includes(pathToPigweed)) {
+  if (
+    pathToPigweed &&
+    !modifiedEnv[pathNameInEnv].split(path.delimiter).includes(pathToPigweed)
+  ) {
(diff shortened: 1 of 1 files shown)


brianignacio5 (changes requested):
We could also validate that `pathToGitDir` and `pathToPigweed` are already in PATH before appending

Fabricio-ESP (approved):
Very specific scenario for this to be an issue, but this is an overall improvement.

PR #1711, opened 2025-11-20
[VSC-1800] - OPENOCD_USB_ADAPTER_LOCATION not being cleared

radurentea:
## Description

This PR introduces support for identifying OpenOCD USB adapters by their serial number (MAC address), improving the reliability of multi-device setups. It also fixes device detection issues by ensuring scans aren't constrained by previous location settings and optimizes OpenOCD version checks.

### OpenOCD Adapter Serial Support
- **Added Serial Number Tracking**: Implemented logic to parse, store, and retrieve the USB adapter's serial number (MAC address) from OpenOCD logs.
- **Target Specificity**: The extension now injects `adapter serial <serial>` into OpenOCD commands when a specific adapter has been identified, taking precedence over `OPENOCD_USB_ADAPTER_LOCATION`.
- **New Module**: Created `src/espIdf/openOcd/adapterSerial.ts` to handle serial number logic.

### Device Detection Improvements
- **Unrestricted Scanning**: Modified `DevkitsCommand` to temporarily unset `OPENOCD_USB_ADAPTER_LOCATION` during device detection. This ensures the scan checks all available devices rather than just the currently configured one.
- **Clean State**: `setIdfTarget` now clears stored adapter serials and locations when selecting a new target to prevent stale configuration issues.
- **Automatic Port Selection**: Setting a connected device target now automatically detects and updates the serial port configuration to match the selected device. This change was made mainly for not having to change port manually before running monitor, which is a common use case for the users.

### Performance
- **Silent Version Checks**: Updated `OpenOCDManager.version()` to support a `silent` mode, reducing log noise during frequent version checks in the hints provider.

## Type of change

1. In a project, you should be able to switch between targets (not connected and connected) without any issues, even if you have two similar boards connected. Also, if the selected target is a connected board, the serial port should also be changed to the correct one. 

## How has this been tested?

Same as described above 

**Test Configuration**:
* ESP-IDF Version: 5.5.1

## Checklist
- [ ] PR Self Reviewed
- [x] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/src/espIdf/setTarget/index.ts
+++ b/src/espIdf/setTarget/index.ts
@@ -196,26 +196,6 @@ export async function setIdfTarget(
             );
             customExtraVars["OPENOCD_USB_ADAPTER_LOCATION"] = location;
           }
-          
-          // Update serial port if board is connected
-          // The serial port should match the connected board
-          try {
-            const detectedPort = await SerialPort.detectDefaultPort(
-              workspaceFolder.uri
-            );
-            if (detectedPort) {
-              await SerialPort.shared().updatePortListStatus(
-                detectedPort,
-                workspaceFolder.uri,
-                false // useMonitorPort = false, update idf.port
-              );
-            }
-          } catch (error) {
-            Logger.info(
-              `Failed to detect serial port for connected board: ${error.message}`,
-              "setIdfTarget"
-            );
-          }
         } else {
           await selectOpenOcdConfigFiles(
             workspaceFolder.uri,


brianignacio5:
The thing about this logic is only applies when user run the setTarget command.

I'm wondering for the case: 1) User has connected board, 2) Run set target, 3) value is set in customExtraVars 4) Time passes user disconnect board 5) User tries to use OpenOCD Server directly and it doesn't work ? USB_ADAPTER_LOCATION is changed ?

Fabricio-ESP (approved):
Confirmed Working.
Thanks for the change.

src/espIdf/openOcd/adapterSerial.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+  return undefined;
+}
+
+/**
+ * Stores the OpenOCD USB adapter serial number in workspace state
+ * @param workspaceFolder The workspace folder URI
+ * @param serialNumber The serial number to store
+ */
+export async function storeAdapterSerial(
```
brianignacio5: This doesn't need to be an async function.

src/espIdf/openOcd/adapterSerial.ts
```diff
@@ -0,0 +1,144 @@
...
+    );
+  }
+  return undefined;
+}
+
+/**
+ * Clears the stored OpenOCD USB adapter serial number from workspace state
+ * @param workspaceFolder The workspace folder URI
+ */
+export async function clearAdapterSerial(
```
brianignacio5: This doesn't need to be an async function.

src/espIdf/openOcd/openOcdManager.ts
```diff
@@ -222,6 +234,19 @@ export class OpenOCDManager extends EventEmitter {
...
       this.sendToOutputChannel(data);
+      
+      // Parse adapter serial number from log output
+      const serialNumber = parseAdapterSerialFromLog(data);
+      if (serialNumber && this.workspace) {
+        storeAdapterSerial(this.workspace, serialNumber).catch((error) => {
+          Logger.error(
+            "Failed to store adapter serial number",
+            error,
+            "OpenOCDManager stderr"
```
brianignacio5: Better to improve the error tag here instead of `OpenOCDManager stderr` to `storeAdapterSerial`

src/espIdf/openOcd/openOcdManager.ts
```diff
@@ -242,6 +267,19 @@ export class OpenOCDManager extends EventEmitter {
...
       this.sendToOutputChannel(data);
+      
+      // Parse adapter serial number from log output
+      const serialNumber = parseAdapterSerialFromLog(data);
+      if (serialNumber && this.workspace) {
+        storeAdapterSerial(this.workspace, serialNumber).catch((error) => {
+          Logger.error(
+            "Failed to store adapter serial number",
+            error,
+            "OpenOCDManager stdout"
```
brianignacio5: Same here: Better use a different tag that explain that the error is related to parsing serial from OpenOCD log

src/espIdf/setTarget/index.ts
```diff
@@ -169,23 +188,33 @@ export async function setIdfTarget(
...
-              configurationTarget,
+          }
+          
+          // Update serial port if board is connected
+          // The serial port should match the connected board
+          try {
+            const detectedPort = await SerialPort.detectDefaultPort(
               workspaceFolder.uri
             );
+            if (detectedPort) {
```
brianignacio5: I can see that we are using detectDefaultPort here to get the first chip that matches selected IDF_TARGET.

This doesn't mean that the first chip detected is the same as the one from `OPENOCD_USB_ADAPTER_LOCATION`. This could be true for only one of same chip connected, but **what happens when we have multiple esp32 connected at the same time?**

Can the serial port (for example `/dev/cu*` be extracted from `selectedTarget.boardInfo` directly ? If yes, we can just update the idf.port setting directly.

if we can't get the serial port from `selectedTarget.boardInfo`, we need to find a way to map `selectedTarget.boardInfo.location` to one of serial ports given by `SerialPort.shared().list()` since I think `SerialPort.detectDefaultPort` will not necessarily return the right port always.

src/espIdf/openOcd/openOcdManager.ts
radurentea: Because I will make the above functions to be sync, I will remove the .catch()

src/espIdf/openOcd/openOcdManager.ts
radurentea: Because I will make the above functions to be sync, I will remove the .catch()

src/espIdf/setTarget/index.ts
radurentea: I can confirm that we did not have the information in `selectedTarget.boardInfo`. I've managed to find a way to correlate the USB adapter location with information from selectedTarget.boardInfo, but the solution is different for each OS.

radurentea: I've removed the functionality of correlating the serial port with the usb adapter location. Changing the target, won't do anything to the serial port settings anymore. (Back to how it was before this PR)

brianignacio5 (approved):
LGTM

PR #1822, opened 2026-04-21
Refine EIM launch detection and remove split GUI/CLI commands

### LINKED ISSUE #1806: eim gui fails to start under certain conditions
### OS

MacOS

### Operating System version

macOS 15.7.1

1.112.0

### ESP-IDF version

any ESP-IDF version

### Python version

-

### Doctor command output

-

### Extension

_No response_

### Description

If you are a macOS user and have installed both eim-gui and eim-cli via Homebrew, you will not be able to launch the eim-gui installation page via the extension. 

A terminal window pops up, but after executing the command, nothing happens—the EIM GUI doesn't appear.

Image

### Debug Message

-

### Other Steps to Reproduce

1. brew uninstall eim
2. The eim GUI opens normally when you launch the 'Open ESP-IDF Installation Manager'
3. brew install eim
4. Unable to open the eim GUI via 'Open ESP-IDF Installation Manager'

radurentea:
## Description

Improve how the extension resolves, installs, and launches EIM.

This change removes the separate GUI and Terminal commands and keeps a single `ESP-IDF: Open ESP-IDF Installation Manager` entry. The extension now prefers an existing `eim` found in `PATH`, detects whether that EIM supports GUI by checking `help-json`, and launches either GUI or CLI automatically.

This change removes the separate GUI and Terminal commands and keeps a single ESP-IDF: Open ESP-IDF Installation Manager entry. The extension now prefers an existing EIM installation, checking PATH first, detects whether that EIM supports GUI by checking help-json, and launches either GUI or CLI automatically.

It also removes the redundant HEAD request during download, makes downloads atomic by writing to a temporary file before moving it into place, fixes macOS handling so CLI-only eim binaries are executed directly instead of being incorrectly opened as app bundles, restores the Snap-specific warning flow for GUI-capable EIM installs, and makes CLI asset selection/path resolution platform-aware for Linux, macOS, and Windows.

Documentation and localization were updated to reflect the new automatic launch behavior.

Fixes #1806

## Type of change

1. Open Command Palette.
2. Run `ESP-IDF: Open ESP-IDF Installation Manager` on a local machine with no existing EIM installation.
3. Select a download mirror if prompted.
4. Observe that the extension downloads and launches the GUI-capable EIM automatically.
5. Install or place a CLI-only `eim` in `PATH`.
6. Run `ESP-IDF: Open ESP-IDF Installation Manager` again.
7. Observe that the extension detects the existing CLI-only EIM and launches it in terminal mode instead of trying to open a GUI.
8. On macOS, verify that a CLI-only Homebrew-installed `eim` is executed directly and does not fail through `open`.
9. In a remote/headless environment (for example WSL, SSH, dev container, Codespaces, or web UI), run `ESP-IDF: Open ESP-IDF Installation Manager`.
10. Observe that the extension installs or launches the CLI workflow automatically.
11. On Linux with VS Code installed via Snap and a GUI-capable EIM available, run `ESP-IDF: Open ESP-IDF Installation Manager`.
12. Observe that the extension shows the Snap warning dialog instead of trying to launch the GUI directly.
13. Choose `Run EIM in Terminal` and verify that EIM launches in CLI mode in the integrated terminal.
14. Repeat the Snap scenario and choose `Copy EIM Path`, then verify that the EIM path is copied to the clipboard.
15. On Snap with no EIM installed, run `ESP-IDF: Open ESP-IDF Installation Manager` and verify that the extension uses the CLI installation/launch path instead of attempting to open the GUI.
16. Confirm that the old `Open ESP-IDF Installation Manager (GUI)` and `Open ESP-IDF Installation Manager (Terminal)` commands are no longer present.
17. Start an EIM download, cancel it before completion, then run ESP-IDF: Open ESP-IDF Installation Manager again.
Observe that no partial download is reused and the extension retries cleanly.
18. Review docs/settings text related to EIM and confirm they match the new behavior.

- Expected behaviour:

  The extension should use a single Install Manager command, prefer an existing `eim` from `PATH`, detect GUI capability automatically, and choose the correct launch or install mode for local, Snap-constrained, and headless environments.

- Expected output:

  EIM launches successfully in GUI mode when supported, falls back to CLI mode when GUI is unavailable, shows the Snap-specific dialog for GUI-capable EIM on Linux Snap installs, uses CLI directly when Snap cannot launch GUI, macOS CLI-only binaries run without `open` errors, and the outdated commands/docs are no longer shown.

## How has this been tested?

As described above

**Test Configuration**:
* ESP-IDF Version: 6.0.0
* OS (Windows,Linux and macOS): All

## Checklist
- [ ] PR Self Reviewed
- [x] Applied Code formatting
- [x] Added Documentation
- [ ] Added Unit Test
- [x] Verified on all platforms - Windows,Linux and macOS

## Summary by CodeRabbit

* **New Features**
  * EIM now auto-detects GUI support and launches GUI or CLI without prompting; installer fetches GUI builds for local desktops and CLI builds for remote/headless environments, and updates launch args automatically.

* **Documentation**
  * Docs updated to describe automatic launch selection, the updated default launch args, and revised snap guidance.

* **Localization**
  * Removed GUI/CLI launch-selection strings from multiple locales.

* **Chores**
  * Removed separate GUI/CLI "Open Installation Manager" commands and related labels.


src/eim/downloadInstall.ts
```diff
@@ -371,41 +489,68 @@ export async function downloadAndInstallEIM(
...
+        });
+        const totalSize = Number.parseInt(
+          fileResponseStream.headers["content-length"] || "0",
+          10
+        );
 
-      let downloadedSize = 0;
-      fileResponseStream.data.on("data", (chunk: Buffer) => {
+        const cancellationError = new Error("Download canceled by user.");
+        cancellationListener = cancelToken.onCancellationRequested(() => {
```
brianignacio5, line 485: cancelToken.onCancellationRequested and if cancelToken.isCancellationRequested seems to be doing the same and reactivate to same event. 

Should we use one or the other ?

src/eim/downloadInstall.ts
```diff
@@ -57,6 +65,116 @@ export function shouldForceCliMode(): boolean {
...
+  }
+
+  if (arch === "x64") {
+    return "x64";
+  }
+
+  throw new Error(`Unsupported architecture: ${arch}`);
+}
+
+function getLinuxCliAssetArch(arch: string): "aarch64" | "armv7" | "x64" {
```
brianignacio5, line 112: getLinuxCliAssetArch and getGuiAssetArch can be just one function.

src/eim/downloadInstall.ts
```diff
@@ -57,6 +65,116 @@ export function shouldForceCliMode(): boolean {
...
+  }
+
+  if (arch === "x64") {
+    return "x64";
+  }
+
+  throw new Error(`Unsupported architecture: ${arch}`);
+}
+
+function getCliAssetName(arch: string): string {
```
brianignacio5: By using a variable = "cli" || "gui" ,the functions getCliAssetName and getGuiAssetName can be one function.

src/eim/downloadInstall.ts
```diff
@@ -57,6 +65,116 @@ export function shouldForceCliMode(): boolean {
...
+      : process.env.HOME || process.env.USERPROFILE);
+
+  if (!homeDir) {
+    throw new Error("Unable to resolve the user home directory.");
+  }
+
+  return homeDir;
+}
+
+function getCliInstallDir(): string {
```
brianignacio5: Same idea as "cli" and "gui" we can see that getCliInstallDir and getGuiInstallDir can become a single function.

src/eim/downloadInstall.ts
```diff
@@ -57,6 +65,116 @@ export function shouldForceCliMode(): boolean {
...
+function getCliBinaryPath(): string {
+  if (process.platform === "win32") {
+    return join(getCliInstallDir(), "eim-cli-windows-x64.exe");
+  }
+
+  return join(getCliInstallDir(), "eim");
+}
+
+function getGuiAssetArch(arch: string): "aarch64" | "x64" {
+  if (arch === "arm64") {
```
brianignacio5: There is a way to implement this without so much space or if use.

```
function getGuiAssetArch(arch: string): "aarch64" | "x64" {
  switch (arch) {
    case "arm64":
      return "aarch64";
    case "x64":
      return "x64";
    default:
      throw new Error(`Unsupported architecture: ${arch}`);
  }
}
```

brianignacio5 (changes requested):
Code improvements I suggest.

src/eim/downloadInstall.ts
radurentea: Actually they handle different cases. Imagine the user clicks cancel while we're still waiting for axios to respond — at that point the listener isn't registered yet, so the cancellation event just gets lost. The isCancellationRequested check catches exactly that situation after axios returns.

So:

`onCancellationRequested` -> user cancels during the download stream
`isCancellationRequested` -> user cancelled while we were waiting for axios

You're right though that the destroy logic is duplicated in both places, which makes it look like the same thing. I'll refactor it to just throw early in the isCancellationRequested block — pipeline hasn't started at that point so no need to destroy anything.

src/eim/downloadInstall.ts
radurentea: I thought about this but I think keeping them separate is actually better. The return types are different on purpose, GUI doesn't support armv7, CLI does. If we merge them we either lose that or we need overloads or some flag parameter, which is more complexity for not much gain. The duplication is just two lines so I'd leave it as is.

src/eim/downloadInstall.ts
```diff
@@ -79,24 +197,22 @@ export async function resolveEimPath(): Promise<string> {
...
+    if (await pathExists(cliPath)) {
+      eimPath = cliPath;
+    }
+  }
+  // 5. Use default GUI path based on platform if still not found
+  if (!eimPath) {
+    const guiPath = getEimBinaryPath(getGuiInstallDir(), false);
+    if (await pathExists(guiPath)) {
+      eimPath = guiPath;
     }
```
kolipakakondal, line 189: For better visibilty and troubleshooting purpose:
1. Document this order which you're checking so that users could check if something goes wrong
2. Log the path you're checking for each step and it helps us in troubleshooting when we ask for doctor command file.

src/eim/downloadInstall.ts
```diff
@@ -79,24 +197,22 @@ export async function resolveEimPath(): Promise<string> {
...
+    const cliPath = getCliBinaryPath();
+    if (await pathExists(cliPath)) {
+      eimPath = cliPath;
+    }
+  }
+  // 5. Use default GUI path based on platform if still not found
+  if (!eimPath) {
+    const guiPath = getEimBinaryPath(getGuiInstallDir(), false);
+    if (await pathExists(guiPath)) {
+      eimPath = guiPath;
```
kolipakakondal: Do we check for GUI after CLI?

src/eim/downloadInstall.ts
kolipakakondal: If there are launch failures - will it be visible to the user?

src/eim/downloadInstall.ts
radurentea: Yes, by mistake. I've made a new commit for it.

src/eim/downloadInstall.ts
radurentea: I've manged to test this and in the case of the "GUI" scenario, it won't, but for the "cli" version, the terminal remains open and it shows the error:

Here is an example of output when I've created a fake file for the eim executable on windows.

```powershell
PS C:\fake> & 'C:\fake\eim.exe' wizard --idf-features ide
Program 'eim.exe' failed to run: The specified executable is not a valid application for this OS platform.At line:1 char:1
+ & 'C:\fake\eim.exe' wizard --idf-features ide
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~.
At line:1 char:1
+ & 'C:\fake\eim.exe' wizard --idf-features ide
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ResourceUnavailable: (:) [], ApplicationFailedException
    + FullyQualifiedErrorId : NativeCommandFailed
```

I've pushed a commit, I couldn't find a better solution than the one in which we keep the terminal open no matter what. Here is an example of how it would look on a successful launching of eim gui
image

src/eim/downloadInstall.ts
radurentea: I agree, that's a very good idea, thanks!

I've implemented it in commit [feat: log resolveEimPath resolution steps and document lookup order](https://github.com/espressif/vscode-esp-idf-extension/pull/1822/commits/009f8f8543a5936d9038001b0f09b9e5d14e948e)

src/eim/downloadInstall.ts
```diff
@@ -57,18 +65,109 @@ export function shouldForceCliMode(): boolean {
...
+  if (process.platform === "win32") {
+    return join(getEimHomeDir(), ".espressif", "eim_gui");
+  }
+
+  if (process.platform === "darwin") {
+    return "/Applications";
+  }
+
+  if (process.platform === "linux") {
+    return join(process.env.HOME || "", ".espressif", "eim_gui");
```
brianignacio5: Why does "linux" needs to use `process.env.HOME` instead of using `getEimHomeDir` ?

Seems to me that this could be further reduced to:

```
function getEimInstallDir(mode: "cli" | "gui"): string {
  if (process.platform === "darwin" && mode === "gui") {
    return "/Applications";
  }
  if (process.platform !== "win32" && process.platform !== "linux" && process.platform !== "darwin") {
    throw new Error(`Unsupported platform: ${process.platform}`);
  }
  const subdir = mode === "cli" ? "eim" : "eim_gui";
  return join(getEimHomeDir(), ".espressif", subdir);
}
```

brianignacio5 (approved):
LGTM with last refactor suggestion.

AndriiFilippov:
@radurentea hi ! 

Tested under: 
OS: Windows 11 / Linux Ubuntu / Mac arm64 

**Windows 11:**
CLI via **winget** ✅ 
GUI via **winget** - as we discussed in the meeting, winget does not add the GUI to the PATH, so the extension repeatedly downloads the GUI version ⚠️ 
GUI via manual setup ✅ 
CLI via manual setup - we don’t support this CLI installation method in the extension yet 🔴 
and it’s a bit confusing since the official EIM web provides this option: 
image

So, does it make sense to add a setting in VSCode to let user set custom path for EIM? For the scenarios in which users download EIM via direct link❓  @kolipakakondal @brianignacio5 
 

**Linux Ubuntu:**
CLI via APT (for Snap-installed VS Code) ✅ 
CLI via APT ✅ 
GUI via APT (for Snap-installed VS Code) - extension provides pop-up with two options: **Run EIM in Terminal** **/** **Copy EIM Path** , able to launch GUI in CLI mode in the integrated terminal ✅

GUI via APT ✅ 
HEADLESS/REMOTE MODE: the extension detects headless mode and installs only the EIM CLI. ✅
( I haven’t seen the issue related to **“ESP-IDF Setup is not valid:”** while working with WSL. @radurentea , could you please verify this scenario on your side? ⚠️ ) 


**Mac arm64:**
CLI via **homebrew** ✅ 
GUI via **homebrew** ✅ 
GUI via manual setup ✅

AndriiFilippov (approved):

radurentea:
Confirmed: in WSL, running Open ESP-IDF Installation Manager from a non-ESP-IDF project works as expected.
The issue I previously reported was caused by a problem in my local setup.

PR #1842, opened 2026-05-12
[VSC-1688/1718/1748] CMakePresets

radurentea:
## Description
This PR migrates the ESP-IDF extension's project configuration system from the proprietary `esp_idf_project_configuration.json` format to the industry-standard [CMake Presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) format (`CMakePresets.json` / `CMakeUserPresets.json`). Configuration profiles are now expressed as CMake `configurePresets`, with ESP-IDF-specific settings (OpenOCD configs, flash settings, monitor settings, etc.) stored under a typed `vendor` extension key.
The Project Configuration GUI editor and the "ESP-IDF: Open Project Configuration" command are removed. Presets are edited directly in `CMakePresets.json`, with schema-backed validation and autocompletion in the editor, and a new command seeds a starter file so nobody has to write the first preset from scratch.
**Key changes:**
- **Migration**: Existing `esp_idf_project_configuration.json` configurations are automatically detected and migrated to `CMakePresets.json` on first use with a confirmation prompt. The original file is left untouched.
- **`CMakeUserPresets.json` support**: Personal settings you do not want to share (baud rates, local build paths) live in `CMakeUserPresets.json` next to the shared `CMakePresets.json`. Preset names must be unique across both files: CMake refuses to read presets when a name is declared twice, and the extension now behaves the same way, reporting an error that names the presets involved. To personalise a shared preset, add a preset under a different name that `inherits` it.
- **New command, `ESP-IDF: Create Project Configuration`**: Writes `CMakePresets.json` with two buildable starter presets (`default` and `production`) modelled on the ESP-IDF `multi_config` example. Presets you already defined are kept, and a file that fails to parse is reported rather than overwritten.
- **CMake-conformant preset resolution**:
  - Presets marked `hidden` are inheritance bases only and are no longer offered for selection, matching `cmake --list-presets`.
  - The `condition` field is evaluated, so presets that do not apply to the current host are not offered. All CMake condition types are supported.
  - With multiple parents in `inherits`, the earlier parent now wins, which is CMake's rule; previously the later parent did.
  - `displayName` and `description` are no longer inherited, and circular inheritance is detected and reported.
  - Macro expansion covers `${presetName}`, `${hostSystemName}`, `${pathListSep}` and `${dollar}` in addition to `${sourceDir}`.
- **Configuration quick pick**: Presets are listed by `displayName` with their `description` as detail, falling back to the preset name.
- **`idf.py` integration**: The selected preset is passed through as `IDF_PRESET` and via `idf.py --preset`, including for `set-target`.
- **JSON schema**: An ESP-IDF CMakePresets schema (`esp-idf-cmakepresets-schema-v1.json`) is bundled and registered for `CMakePresets.json` / `CMakeUserPresets.json`, providing in-editor validation and autocompletion with no `$schema` field needed in the file. Note that a `$schema` field must **not** be added: it requires CMake Presets version 8 (CMake 3.28) whereas ESP-IDF is on version 3, and CMake refuses to read the entire file if it is present.
- **Generated files target the CMake that ESP-IDF ships**: `cmakeMinimumRequired` is now 3.21.0, the floor for presets version 3. It previously claimed 3.23.0, the floor for version 4, which the CMake 3.22.1 that `idf_tools` installs refuses to read. The bundled schema also no longer requires `cmakeMinimumRequired`, so ESP-IDF's own `multi_config` example validates.
- **CMake extension conflict prevention**: The CMake Tools extension's auto-configure is suppressed when the ESP-IDF extension is active to avoid conflicting preset handling.
- **Startup preset restore**: A new `idf.saveLastProjectConfiguration` setting (enabled by default) makes the extension remember the last selected configuration preset across workspace reopens. When disabled, no preset is pre-selected on startup.
- **Build path consistency**: `idf.buildPath` now always reflects the active preset's `binaryDir`, fixing stale paths after profile edits or target changes.
- **CI schema upload**: A new GitHub Actions workflow publishes the schema file as an upload artefact on release.
- **Documentation**: Rewritten guides for project configuration and multiple projects, plus both new commands added to the command reference in English and Chinese.
## Type of change
**Migration from legacy format:**
1. Open a workspace that has an existing `esp_idf_project_configuration.json` file with one or more configuration profiles.
2. Observe the migration prompt asking whether to convert to CMakePresets format.
3. Accept the migration.
- **Expected behaviour:** A `CMakePresets.json` file is created at the workspace root containing one `configurePreset` entry per legacy profile. The original `esp_idf_project_configuration.json` is left intact.
- **Expected output:** The status bar shows the previously selected profile, and the configuration list shows all migrated presets.
**Creating a configuration from scratch:**
1. In a project with no `CMakePresets.json`, run `ESP-IDF: Create Project Configuration`.
- **Expected behaviour:** `CMakePresets.json` is created with `default` and `production` presets and opened in the editor. `cmake --list-presets` in a terminal lists both.
- **Expected output:** Both presets appear in `ESP-IDF: Select Project Configuration`. Running the command again reports that presets are already defined and changes nothing.
**Selecting a configure preset:**
1. Click the active-configuration badge in the VS Code status bar (or run `ESP-IDF: Select Project Configuration`).
2. Choose a preset from the quick-pick list.
- **Expected behaviour:** All subsequent build/flash/monitor commands use the selected preset's settings (`IDF_TARGET`, OpenOCD configs, build directory, etc.).
- **Expected output:** Presets are listed by `displayName` with `description` shown as detail. The status bar updates to display the newly selected preset name.
**Hidden presets and conditions:**
1. Add a preset with `"hidden": true` and a second preset that lists it in `inherits`.
2. Add a third preset with a `condition` that is false on your host, for example `{"type": "equals", "lhs": "${hostSystemName}", "rhs": "Nonexistent"}`.
3. Open the configuration quick-pick.
- **Expected behaviour:** Neither the hidden preset nor the false-condition preset is offered, matching `cmake --list-presets`. The preset that inherits the hidden one still receives all of its settings.
**Duplicate preset names:**
1. Declare the same preset name in both `CMakePresets.json` and `CMakeUserPresets.json`.
2. Run `ESP-IDF: Select Project Configuration`.
- **Expected behaviour:** No configuration is offered while the duplicate exists, and an error message names the presets involved. `cmake --list-presets` reports an error for the same file.
- **Expected output:** Renaming one of them restores the list.
**`CMakeUserPresets.json`:**
1. Create a `CMakeUserPresets.json` at the workspace root with a `configurePreset` entry under a **different** name that inherits one defined in `CMakePresets.json`, and override a setting such as the monitor baud rate.
2. Open the configuration quick-pick.
- **Expected behaviour:** Both presets appear in the list. Selecting the user preset applies its overrides and inherits everything else from its parent.
- **Expected output:** The extension does not modify `CMakeUserPresets.json` unless you change settings while that preset is selected.
**JSON schema validation:**
1. Open `CMakePresets.json` or `CMakeUserPresets.json` in the editor.
2. Add an unrecognised key inside the `vendor["espressif/vscode-esp-idf"]` object.
- **Expected behaviour:** The editor underlines the invalid key with a schema validation error, without any `$schema` field in the file.
- **Expected output:** Hovering the squiggle shows a descriptive message from `esp-idf-cmakepresets-schema-v1.json`.
**Generated file works with the bundled CMake:**
1. Create a configuration with the new command, then run `cmake --list-presets` using the CMake that `idf_tools` installed.
- **Expected behaviour:** The presets are listed. No "cmakeMinimumRequired" version error.
**Build path consistency after target change:**
1. Select a preset whose `binaryDir` is set to a custom path (e.g. `${sourceDir}/build_custom`).
2. Run `ESP-IDF: Set Espressif Device Target` and change the target.
- **Expected behaviour:** The build directory remains the preset's `binaryDir`.
- **Expected output:** No fallback to the default `${workspaceFolder}/build` path; the build output appears in `build_custom/`.
## How has this been tested?
Unit tests cover preset parameter mapping, legacy migration, inheritance precedence, hidden presets, condition evaluation, duplicate-name rejection and starter-file creation. The generated presets and every JSON example in the documentation were additionally checked against a real `cmake --list-presets` run and against the bundled schema.
**Test Configuration**:
* ESP-IDF Version: 6.0.0
## Checklist
- [ ] PR Self Reviewed
- [x] Applied Code formatting
- [x] Added Documentation
- [x] Added Unit Test


src/project-conf/index.ts
```diff
@@ -311,200 +692,1082 @@ function resolveConfigPaths(
...
+  switch (param) {
+    case "idf.cmakeCompilerArgs":
+      return getESPIDFSettingValue(preset, "compileArgs") || "";
+    case "idf.ninjaArgs":
+      return getESPIDFSettingValue(preset, "ninjaArgs") || "";
+    case "idf.buildPath":
+      return preset.binaryDir || "";
+    case "idf.sdkconfigDefaults":
+      const sdkconfigDefaults = preset.cacheVariables?.SDKCONFIG_DEFAULTS;
+      return sdkconfigDefaults ? sdkconfigDefaults.split(";") : "";
```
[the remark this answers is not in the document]
radurentea: Suggestion is wrong, but for the false case, the value should be an [], not "".

src/espIdf/setTarget/index.ts
```diff
@@ -212,7 +240,11 @@ export async function setIdfTarget(
           configurationTarget,
           workspaceFolder.uri
         );
+
         updateOpenOcdAdapterStatusBarItem(workspaceFolder.uri);
+
+        // Update project configuration with IDF_TARGET if a configuration is selected
+        // Note: IDF_TARGET goes in cacheVariables, not environment
```
brianignacio5: Using the `multi_config` with 3 project configuration 2 just have a custom build directory and 1 a custom sdkconfig file path. After selecting the profile from the status bar I can build the project without issue but when I try to use the `ESP-IDF: Set Espressif Device Target` command I see this error:

```
[Set Target]
Running IDF Set Target action
[0;33mCMake presets file found but no preset name given; using first preset: 'proj1'[0m

CMake Error: Could not read presets from /Users/brainignacio/workspace/tests/v60/multi_config:
Invalid macro expansion in "proj1"

cmake failed with exit code 1, output of the command is in the /Users/brainignacio/workspace/tests/v60/multi_config/build3/log/idf_py_stderr_output_85152 and /Users/brainignacio/workspace/tests/v60/multi_config/build3/log/idf_py_stdout_output_85152

Adding "set-target"'s dependency "fullclean" to list of commands with default set of options.
Executing action: fullclean
Executing action: set-target
Set Target to: esp32c3, new sdkconfig will be created.
Running cmake in directory /Users/brainignacio/workspace/tests/v60/multi_config/build3
Executing "cmake -G Ninja -B /Users/brainignacio/workspace/tests/v60/multi_config/build3 -DPYTHON_DEPS_CHECKED=1 -DPYTHON=/Users/brainignacio/.espressif/tools/python/v6.0.1/venv/bin/python3 -DESP_PLATFORM=1 -DIDF_TARGET=esp32c3 -DCCACHE_ENABLE=False /Users/brainignacio/workspace/tests/v60/multi_config --preset proj1"...

Failed to set target esp32c3: non zero exit code 2

[0;33mCMake presets file found but no preset name given; using first preset: 'proj1'[0m
CMake Error: Could not read presets from /Users/brainignacio/workspace/tests/v60/multi_config:
Invalid macro expansion in "proj1"
cmake failed with exit code 1, output of the command is in the /Users/brainignacio/workspace/tests/v60/multi_config/build3/log/idf_py_stderr_output_85152 and /Users/brainignacio/workspace/tests/v60/multi_config/build3/log/idf_py_stdout_output_85152
Adding "set-target"'s dependency "fullclean" to list of commands with default set of options.
Executing action: fullclean
Executing action: set-target
Set Target to: esp32c3, new sdkconfig will be created.
Running cmake in directory /Users/brainignacio/workspace/tests/v60/multi_config/build3
Executing "cmake -G Ninja -B /Users/brainignacio/workspace/tests/v60/multi_config/build3 -DPYTHON_DEPS_CHECKED=1 -DPYTHON=/Users/brainignacio/.espressif/tools/python/v6.0.1/venv/bin/python3 -DESP_PLATFORM=1 -DIDF_TARGET=esp32c3 -DCCACHE_ENABLE=False /Users/brainignacio/workspace/tests/v60/multi_config --preset proj1"...
```
Here is the CMakePresets file:
[CMakePresets.json](https://github.com/user-attachments/files/29584537/CMakePresets.json)


Here is the output of `/Users/brainignacio/workspace/tests/v60/multi_config/build3/log/idf_py_stderr_output_85152`:

```
Command: cmake -G Ninja -B /Users/brainignacio/workspace/tests/v60/multi_config/build3 -DPYTHON_DEPS_CHECKED=1 -DPYTHON=/Users/brainignacio/.espressif/tools/python/v6.0.1/venv/bin/python3 -DESP_PLATFORM=1 -DIDF_TARGET=esp32c3 -DCCACHE_ENABLE=False /Users/brainignacio/workspace/tests/v60/multi_config --preset proj1
CMake Error: Could not read presets from /Users/brainignacio/workspace/tests/v60/multi_config:
Invalid macro expansion in "proj1"
```
The other file has the same without error message.

If I delete the existing build directory it works. I also tried again with new session and was working ok.

In summary, we need a way to specify the preset in other commands because idf.py command will use preset if CMakePresets.json is present in current ESP-IDF project.

src/project-conf/index.ts
```diff
@@ -67,37 +164,313 @@ export async function updateCurrentProfileIdfTarget(
...
+
+/**
+ * Legacy compatibility function to save project configuration elements to CMakePresets.json.
+ * This function converts legacy ProjectConfElement objects to ConfigurePreset format and saves them.
+ * Used during migration from the old esp_idf_project_configuration.json format.
+ * @param workspaceFolder The workspace folder Uri where the configuration file will be saved
+ * @param projectConfElements An object mapping configuration names to their legacy ProjectConfElement objects
+ * @deprecated Use saveProjectConfFile instead for new code
+ */
+export async function saveProjectConfFileLegacy(
```
brianignacio5: We could move this and legacy support to a legacy.ts file or other file to separate and keep the project-conf clean with new implementation.

src/project-conf/index.ts
```diff
@@ -311,200 +696,1099 @@ function resolveConfigPaths(
...
+ * - Flash and monitor baud rates
+ * 
+ * Each setting is processed to substitute variables in string values, arrays, and nested objects.
+ * @param vendor The vendor settings object from the preset (under "espressif/vscode-esp-idf" key)
+ * @param workspaceFolder The workspace folder Uri for variable substitution
+ * @param preset The ConfigurePreset containing other variables for substitution
+ * @param resolvePaths If true, paths in vendor settings will be resolved to absolute paths
+ * @returns A new vendor settings object with all variables substituted
+ */
+async function processConfigurePresetVendor(
```
brianignacio5: This file has grown quite a bit.

I would suggest splitting the file into smaller files separated by functionality.

Something like getters, setters, process, substitution  etc.

src/project-conf/index.ts
```diff
@@ -67,37 +164,313 @@ export async function updateCurrentProfileIdfTarget(
...
     selectedConfig,
-    projectConfJson[selectedConfig]
+    updatedConfig
   );
-  await saveProjectConfFile(workspaceFolder, projectConfJson);
+
+  // Keep in-memory store consistent with consumers expecting legacy ProjectConfElement
+  // Re-read processed presets (with resolved paths) and convert to legacy shape
+  try {
+    const resolvedConfigs = await getProjectConfigurationElements(
```
brianignacio5: I think if the in-memory store needs to change to adapt to CMakePresets object is fine instead of keeping so much of the legacy code.

I would suggest the migration to be simplified to just copy from old json to CMakePresets and use the generated file with new implementation instead of keeping logic to process previous implementation.

 The previous implementation still needed to select the current profile every time you open a workspace folder so in-memory storage should be able to update to CMakePresets without dealing with old in-memory objects.

brianignacio5 (changes requested):
Thank you for the hard work!

A few points I noticed:
1)  simplify index.ts to smaller more focused files to handle logic and improve readibility.
2) Remove legacy code as much as possible, we need to be able to generate CMakePresets from old json file and use the new logic for the most part.
3) Review how preset is passed to extension commands. (See ESP-IDF Set Espressif Device Target comment with error)

Orfait:
Hello
2 months later, is there any progress here ?
It is a pain to not be able to use CMakePresets directly in the extension

radurentea:
Hi @Orfait ,

I'll be pushing the final changes to this PR today, including the modifications Brian requested.

Once the PR checks have passed, I can ping you. It would be very helpful if you could test the new implementation using the generated .vsix package and let us know if everything works as expected on your end.

Orfait:
Hi @radurentea 
I will test it with pleasure, but I will be on vacation for the first 3 weeks of august. If available, I will do my best on tomorrow.

radurentea:
Hi @Orfait,

Whenever you have time, feel free to test using the .vsix by click menu View -> Command Palette..., type Install from VSIX and then select downloaded esp-idf-extension.vsix file to install the extension.

[esp-idf-extension.vsix.zip](https://nightly.link/espressif/vscode-esp-idf-extension/actions/artifacts/8751876246.zip)

Thank you!

Orfait:
Hi @radurentea
Installed and quickly tested with success.
It is working with 3 configs with some combinations of sdkconfig. Intellisense is also working.

radurentea:
Hi @Orfait, thank you for giving it a try and updating us.

AndriiFilippov:
@radurentea hi ! 

Tested under: 
OS: Windows 11 
ESP-IDF: v6.0.2 

Extension does not filter `hidden` presets at all. 
A configure preset with `"hidden": true` is correctly excluded and rejected by native CMake:
```
(venv) PS C:\Users\andri\CmakePresetsTest> cmake --list-presets                                          
Pressing Enter to exit...
Available configure presets:

  "esp32c6-local" - Local ESP32-C6
  "esp32c6"       - Shared ESP32-C6
  ```
  
   but remains visible and selectable in `ESP-IDF: Select Project Configuration` :
   
image

To reproduce: create project, add these files to project:
[CMakeUserPresets.json.txt](https://github.com/user-attachments/files/30588302/CMakeUserPresets.json.txt)
[CMakePresets.json.txt](https://github.com/user-attachments/files/30588304/CMakePresets.json.txt)

 root and verify: 
from terminal `cmake --list-presets`
run command `ESP-IDF: Select Project Configuration`

**Comment by AndriiFilippov** [comment]
Did migration from old project with `esp_idf_project_configuration.json` . When you open PR - there is a notification message asking about legacy migration from `esp_idf_project_configuration.json` to `CMakePresets.json` , but it disappears after 5 seconds.  This is very important migration. **Notification should not disappears without user intervention**. So the notification should be present untill closed by user.

**Comment by AndriiFilippov** [comment]
RP removes `ESP-IDF: Open Project Configuration` command which allowed users to create Presets using pattern. 
Removing the editor is understandable after migrating to CMake Presets, but the PR removes **the only GUI workflow** for creating and editing project configurations without providing an equivalent replacement. Now users **MUST now manually author** a relatively complex CMakePresets.json structure, including ESP-IDF vendor settings. At minimum, the extension should provide a command to generate a valid starter preset pattern or a simplified preset creation wizard.
It is also correlate with Q3 goals.

**Comment by AndriiFilippov** [comment]
will continue on Monday

**Comment by AndriiFilippov** [comment]
OS: Windows 11 
ESP-IDF: v6.0.2

Multiple inheritance precedence is reversed.
The extension gives the later parent higher priority, while CMake gives the earlier parent higher priority.


**to reproduce:** 
have only CMakePresets.json with following content: 
[CMakePresets.json.txt](https://github.com/user-attachments/files/30656564/CMakePresets.json.txt)

add `message(STATUS "QA_PARENT_WINNER=$ENV{QA_PARENT_WINNER}")` to root CMakeLists.txt

For this test, use native CMake, because CMake defines the inherits rules.
**If there is a conflict, the earlier parent in the inherits array should take precedence.**
For a preset declared with:
`"inherits": ["first-parent", "second-parent"]`
must win `first-parent`.

1. running  `cmake --preset precedence-child` from cli creates `build\from-first`  ✅ 

2. **BUT** running `Build` command from the extension resolved presets like this:
```
Selected configuration: precedence-child
Build directory: build/from-second
IDF_TARGET: esp32c6
SDKCONFIG: build/from-second/sdkconfig
QA_PARENT_WINNER: SECOND
```


[Doctor_command.txt](https://github.com/user-attachments/files/30656994/Doctor_command.txt)

**Comment by AndriiFilippov** [comment]
The current `ESP-IDF: Select Project Configuration` behavior turns each configurePresets element into a selectable configuration, without filtering `hidden` or evaluating `condition` . Quick Pick is built from all received keys.

to reproduce: 
create OS specific presets in CMakePresets.json: 
[CMakePresets.json.txt](https://github.com/user-attachments/files/30657382/CMakePresets.json.txt)
file contains different OS presets `windows-only` , `linux-only`. 

Condition should define if the preset is active or disabled on this pacticular OS:
running from extension terminal from Windows 11 : `cmake --list-presets` I do see it resolves `conditions` correctly. 
```
(venv) PS C:\Users\andri\CmakePresetsTest> cmake --list-presets                                          
Available configure presets:            
                          
  "windows-only"    - Windows only
  "child-of-hidden" - Child of hidden base
  ```
  
  Running `ESP-IDF: Select Project Configuration` lets me select any: 
  
image

AndriiFilippov:
Same for duplicated names.
if there is a duplicated names in `CMakePresets.json` and `CMakeUserPresets.json` 
[CMakeUserPresets.json.txt](https://github.com/user-attachments/files/30657640/CMakeUserPresets.json.txt)
[CMakePresets.json.txt](https://github.com/user-attachments/files/30657642/CMakePresets.json.txt)

the cmake command from cli gives you an error:
```
(venv) PS C:\Users\andri\CmakePresetsTest> cmake --list-presets       
CMake Error: Could not read presets from C:/Users/andri/CmakePresetsTest:
Duplicate preset: "debug"
```
but `ESP-IDF: Select Project Configuration` lets you pick duplicated configuration.

AndriiFilippov:
@radurentea hi ! 
Tested under: 
OS: Windows 11
ESP-IDF: v6.0.2 

All issues are fixed now ✅ 

* `hidden` presets are no longer shown in **ESP-IDF: Select Project Configuration** 
* preset `condition` is respected
* multiple inheritance precedence now matches native CMake behavior
* duplicate preset names across `CMakePresets.json` and `CMakeUserPresets.json` are handled correctly
* legacy migration flow works correctly
* project configuration creation/editing flow has been improved by adding `ESP-IDF: Create Project Configuration` command, after removal of the old editor 

LGTM 👍

AndriiFilippov (approved):

brianignacio5 (changes requested):
1. zh_CN docs out of sync
only
- docs_espressif/zh_CN/commands.rst
- docs_espressif/zh_CN/settings.rst updated. 

Need to update 
- docs_espressif/zh_CN/additionalfeatures/project-configuration.rst
- docs_espressif/zh_CN/additionalfeatures/multiple-projects.rst
since still describe the old GUI editor and `esp_idf_project_configuration.json`.
2. `param.saveLastProjectConfiguration` only in English package.nls.json — missing from package.nls.es/pt/ru/zh-cn.json.
3. Update `project-configuration.rst` title from `Project Configuration Editor` to new title.
4. **Schema being used is not in trusted location.** We need to either add domains to `json.schemaDownload.trustedDomains` vscode configuration setting (Required domains are `dl.espressif.com` and `https://raw.githubusercontent.com/Kitware/CMake/`) or have both schema locally in the extension and refer them. This means that we need to update `"$ref": "https://raw.githubusercontent.com/Kitware/CMake/v3.21.0` to a local json schema file `"$ref": "./cmake-presets-schema-v3.21.0.json"` and remove `$id` .
6. **Using this ESP-IDF CMake presets assumes that idf.py supports `IDF_PRESET` (for build, flash, etc.) but this is true only for ESP-IDF v6.x and higher. We need to block this feature entirely for v5.x and lower**

Otherwise LGTM

AndriiFilippov:
LGMT

brianignacio5 (approved):
LGTM

PR #1849, opened 2026-05-25
Run UI Tests on Self-Hosted Runner

AndriiFilippov:
## Description

1. Separated hardware dependent tests (Flash test) from other UI tests. 
2. Added ui-test-helpers.ts to avoid the code duplication. 

## Type of change

1. Click on "Command"
4. Execute action.
5. Observe results.

- Expected behaviour:

- Expected output:

## How has this been tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- Test A
- Test B

**Test Configuration**:
* ESP-IDF Version:
* OS (Windows,Linux and macOS):

## Dependent components impacted by this PR:

- Component 1
- Component 2

## Checklist
- [X] PR Self Reviewed

## Summary by CodeRabbit

* **Chores**
  * Improved end-to-end test workflow: moved UI hardware tests to a dedicated runner, streamlined test execution, ensured hardware test results and screenshots are always published, and aligned the tooling version used during setup.
* **Tests**
  * Added a UI/integration test covering project creation, build and flash flow with device selection and success verification, plus helper utilities to automate opening and driving the test workspace.

### DIFF
--- a/src/ui-test/ui-test-helpers.ts
+++ b/src/ui-test/ui-test-helpers.ts
@@ -1,0 +1,37 @@
+/*
+ * Project: ESP-IDF VSCode Extension
+ * Copyright 2026 Espressif Systems (Shanghai) CO LTD
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { resolve } from "path";
+import { InputBox, Workbench } from "vscode-extension-tester";
+
+export const testWorkspaceDir = resolve(
+  __dirname,
+  "..",
+  "..",
+  "testFiles",
+  "testWorkspace"
+);
+
+export async function openTestProject(): Promise<void> {
+  await new Promise((res) => setTimeout(res, 5000));
+  await new Workbench().executeCommand("file: open folder");
+  await new Promise((res) => setTimeout(res, 1000));
+  const input = await InputBox.create();
+  await input.setText(testWorkspaceDir);
+  await input.confirm();
+  await new Promise((res) => setTimeout(res, 4000));
+}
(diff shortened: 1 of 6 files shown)


brianignacio5 (approved):
LGTM

But I suggest is better if we split the UI Tests in 2 parts for better usage in the future:

1) UI Tests that doesn't rely on real hardware can be run on Github `ubuntu-latest`
2) UI Tests that rely on hardware like desired flash monitor debug E2E tests.

This way we limit usage of self hosted runner to tests (2) and keep (1) on Github.

What do you think ?

AndriiFilippov:
> LGTM
> 
> But I suggest is better if we split the UI Tests in 2 parts for better usage in the future:
> 
> 1. UI Tests that doesn't rely on real hardware can be run on Github `ubuntu-latest`
> 2. UI Tests that rely on hardware like desired flash monitor debug E2E tests.
> 
> This way we limit usage of self hosted runner to tests (2) and keep (1) on Github.
> 
> What do you think ?

I agree. Our Linux self-hosted runner may become a scarce and fragile resource, especially since it will also be shared with Eclipse UI tests.

AndriiFilippov:
> LGTM
> 
> But I suggest is better if we split the UI Tests in 2 parts for better usage in the future:
> 
> 1. UI Tests that doesn't rely on real hardware can be run on Github `ubuntu-latest`
> 2. UI Tests that rely on hardware like desired flash monitor debug E2E tests.
> 
> This way we limit usage of self hosted runner to tests (2) and keep (1) on Github.
> 
> What do you think ?

Hi @brianignacio5 
Please, review

src/ui-test/ui-test-helpers.ts
```diff
@@ -0,0 +1,134 @@
...
+export const testWorkspaceDir = resolve(
+  __dirname,
+  "..",
+  "..",
+  "testFiles",
+  "testWorkspace"
+);
+
+/** Must match command palette labels exactly (category + package.nls title). */
+export const ESP_IDF_COMMANDS = {
```
brianignacio5, line 30: I don't know if possible but maybe you could probably use the command ID from package.json instead of having to keep track of the commands nls title here. I believe it should work the same, but you can test this to see if it makes sense.

These commands id are either in package.json or https://github.com/espressif/vscode-esp-idf-extension/blob/master/src/cmdTreeView/cmdStore.ts#L34

brianignacio5 (approved):
LGTM with small suggestion.

src/ui-test/ui-test-helpers.ts
AndriiFilippov: @brianignacio5 hi ! 
I did use command ID but the way `executeCommand` works is - it select the first match from the top of the list. 
image
so, I decided to use nls title here.

PR #1914, opened 2026-08-07
add gdbinit files in debug config resolution

### LINKED ISSUE #1824: Debugging cannot find bounds of current function (VSC-1913)
### OS

Linux

### Operating System version

Ubuntu 24.04.4 LTS

1.117.0

### ESP-IDF version

2.0.2

### Python version

3.12.3

### Doctor command output

[report.txt](https://github.com/user-attachments/files/26995265/report.txt)

### Extension

[esp_idf_vsc_ext.log](https://github.com/user-attachments/files/26995271/esp_idf_vsc_ext.log)

### Description

When stepping through code, either with "Step" or "Step Into", I hit an error where the call stack for the main (and only) task shows "?? Unknown Source (0)". At this point, the highlighted code goes back to the previous frame and I get the error notification "Selected thread is running". Trying to step again gives the warning "Cannot find bounds of current function".

This problem occurs no matter where I set the breakpoint in my program; after a few steps, this will occur. I can debug 'normally' by just setting breakpoints manually. I only have one breakpoint set. I can restart debugging without having to shut down openOCD.

I am using anESP32-S3-WROOM-1 via USB. My vscode config is added below. I am using one of mutiple project configurations.

Image

[.vscode.zip](https://github.com/user-attachments/files/26995886/default.vscode.zip)

### Debug Message

GDB Debug Console:  [gdb_debug_console.txt](https://github.com/user-attachments/files/26995592/gdb_debug_console.txt)

Open OCD Output (level 3): [openocd_log.txt](https://github.com/user-attachments/files/26995602/openocd_log.txt)

Open OCD Output from step that breaks things: [openocd_debug2.txt](https://github.com/user-attachments/files/26995957/openocd_debug2.txt)

### Other Steps to Reproduce

_No response_

brianignacio5:
## Description

This pull request introduces several improvements and bug fixes to the ESP-IDF VS Code extension, focusing on enhanced debugging configuration, improved error handling during debug sessions, and better support for additional ESP32 targets. The changes also include updates to documentation and minor code cleanups.

**Debugging and Configuration Improvements:**

* Refactored the logic for setting up GDB initialization commands in `CDTDebugConfigurationProvider`, ensuring that prefix map files are sourced if present and providing user feedback if expected files are missing when reproducible builds are enabled. This improves the reliability and clarity of debug configuration setup. [[1]](diffhunk://#diff-e2fa8de3075b5006c67694fcd088159136af77c6f12fdc3dbecf80de0485d0d1L119-R174) [[2]](diffhunk://#diff-e2fa8de3075b5006c67694fcd088159136af77c6f12fdc3dbecf80de0485d0d1R237-R252)
* Expanded the list of supported ESP32 targets and updated the mapping for CPU watchpoint numbers, ensuring correct GDB configuration per target. The placeholder `{IDF_TARGET_CPU_WATCHPOINT_NUM}` (and its braced form) is now correctly replaced in all relevant commands. [[1]](diffhunk://#diff-e2fa8de3075b5006c67694fcd088159136af77c6f12fdc3dbecf80de0485d0d1R188-R220) [[2]](diffhunk://#diff-e7cd97d844562b85fa216d530b6c17ba5ef1196fa5445ffbc240099a8dc30818L161-R161) [[3]](diffhunk://#diff-d6a5b8e6a52479da7d88eda8b06835cc0fbfb07adcae0910ee48f84db2978930L161-R161)

**Debug Session Error Handling:**

* Enhanced error handling in `GDBTargetDebugSession`'s disconnect logic to gracefully handle failures when closing the serial port, disconnecting from the remote target, exiting GDB, or stopping the GDB server. Errors are now reported to the user via output events, but the disconnect process always completes to avoid hanging the session.
* Added logic to detect when OpenOCD stops unexpectedly and automatically end the associated debug session, notifying the user with a warning. [[1]](diffhunk://#diff-52f3a37c5461f3a9923373871914ea79c5405c2d2f75b504bbf269642186fa53R373) [[2]](diffhunk://#diff-04bba6a35cad1c794cbbe677678a51de13441b7a6ee8592b7b50be1f05c6f626R626-R643)

**Other Improvements and Cleanups:**

* Improved merging and writing of EIM executable arguments in both the EIM launch and notification flows, ensuring that duplicate or conflicting arguments are filtered out and the configuration is updated correctly. [[1]](diffhunk://#diff-04bba6a35cad1c794cbbe677678a51de13441b7a6ee8592b7b50be1f05c6f626L4478-R4508) [[2]](diffhunk://#diff-04bba6a35cad1c794cbbe677678a51de13441b7a6ee8592b7b50be1f05c6f626L4505-R4544)
* Minor code cleanups for clarity and consistency, such as more robust parsing of content-length in downloads and improved path handling in Python environment detection. [[1]](diffhunk://#diff-32bdf151a07d1ec0928b1bbbbe6413a780bbbb5fd0aa4c7890a85f4a38f9b6d9L480-R480) [[2]](diffhunk://#diff-0ccfa2b5ac8958888ebfcbcefc4f4ee7086002952fd91deb743a28470bbff72eL124-R142)

**Documentation Updates:**

* Updated both English and Chinese documentation to clarify that the `IDF_TARGET_CPU_WATCHPOINT_NUM` placeholder (including its braced form) is resolved for both default and user-defined `initCommands`. [[1]](diffhunk://#diff-e7cd97d844562b85fa216d530b6c17ba5ef1196fa5445ffbc240099a8dc30818L161-R161) [[2]](diffhunk://#diff-d6a5b8e6a52479da7d88eda8b06835cc0fbfb07adcae0910ee48f84db2978930L161-R161)

Fixes #1824 

## Type of change

1. Press F5 to start a debug session.
2. Check the Debug Console. See that symbols and prefix_map are added to the debug session even before connection.
3. During a debug session, stop the OpenOCD Server from running. The debug session should be terminated as well.
4. Observe results.

- Expected behaviour:

Symbols are added to the debug session automatically.

- Expected output:

## How has this been tested?

Manually using steps above.

**Test Configuration**:
* ESP-IDF Version: 5.5.4

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

## Summary by CodeRabbit

- **Documentation**
  - Clarified debugger command execution order and connection behavior.
  - Updated launch and Native Debug examples with simplified configurations and standard remote-target commands.
  - Removed the obsolete `debugPort` setting and clarified initial breakpoint behavior.

- **Bug Fixes**
  - Improved debugger disconnection cleanup and error handling.
  - Prevented unnecessary initial breakpoints when none are configured.
  - Improved handling and reporting when OpenOCD sessions close unexpectedly.
  - Added localized OpenOCD status and message text.

### DIFF
--- a/docs_espressif/en/debugproject.rst
+++ b/docs_espressif/en/debugproject.rst
@@ -153,12 +153,26 @@ You can modify the configuration to suit your needs. Let's describe the configur
 
 - ``type``: The type of the debug configuration. It should be set to ``gdbtarget``.
 - ``program``: ELF file of your project build directory to execute the debug session. You can use the command ``${command:espIdf.getProjectName}`` to query the extension to find the current build directory project name.
-- ``initCommands``: GDB Commands to initialize GDB and target. The default value is ``["set remote hardware-watchpoint-limit IDF_TARGET_CPU_WATCHPOINT_NUM", "mon reset halt", "maintenance flush register-cache"]``.
-- ``initialBreakpoint``: When ``initCommands`` is not defined, this command will add to default ``initCommands`` a hardware breakpoint at the given function name. For example app_main, the default value, will add ``thb app_main`` to default initCommmands. If set to "", an empty string, no initial breakpoint will be set and if let undefined it will use the default thb app_main.
+- ``initCommands``: Optional GDB commands run **after** the target is connected via ``target.connectCommands``. The extension does not inject a default ``initCommands`` list.
+- ``initialBreakpoint``: When set to a non-empty string (for example ``app_main``), the extension appends ``thb <value>`` to ``initCommands``. If set to ``""`` (empty string) or omitted, no breakpoint is added from this setting.
 - ``gdb``: GDB executable to be used. By default "${command:espIdf.getToolchainGdb}" will query the extension to find the ESP-IDF toolchain GDB for the current IDF_TARGET of your esp-idf project (esp32, esp32c6, etc.).
 
 .. note::
-     ``IDF_TARGET_CPU_WATCHPOINT_NUM`` is resolved by the extension from ``CONFIG_SOC_CPU_WATCHPOINTS_NUM`` in your project's ``sdkconfig``, for both default and user-defined ``initCommands``. If the value is missing, it falls back to ``2``. The braced form of this placeholder is also supported.
+     **GDB command order.** The Eclipse CDT GDB Adapter runs ``target.connectCommands`` first to attach to the target, then runs ``initCommands``.
+
+     When files exist under ``idf.buildPath``, the extension **prepends** commands to ``target.connectCommands`` (before any user-defined ``connectCommands``), in this order:
+
+     1. ``gdbinit/gdbinit`` — ESP-IDF aggregate script (typically sources symbols and connect; available from ESP-IDF v5.3.3). If this file is missing, the extension prepends the same connect sequence as ESP-IDF's ``gdbinit/connect``:
+
+        - ``set remotetimeout 10``
+        - ``target remote :3333``
+        - ``monitor reset halt``
+        - ``maintenance flush register-cache``
+        - ``thbreak app_main``
+
+     2. ``gdbinit/prefix_map``, or fallback ``prefix_map_gdbinit`` — path remapping for reproducible builds. This is a separate file and is **not** included inside ``gdbinit/gdbinit``.
+
+     On ESP-IDF ≥ 5.3.3, ``build/gdbinit/gdbinit`` usually already connects to OpenOCD, so extra user ``connectCommands`` are optional and additive. If ``CONFIG_APP_REPRODUCIBLE_BUILD`` is enabled and no prefix map file is found, the extension shows an information message.
 
 Some additional arguments you might use are:
 
@@ -200,11 +214,11 @@ Some additional arguments you might use are:
             "host": "Target host to connect to (defaults to 'localhost', ignored if parameters is set)",
             "port": "Target port to connect to (defaults to value captured by serverPortRegExp, ignored if parameters is set)",
             "parameters": "Target parameters for the type of target. Normally something like localhost:12345. (defaults to `${host}:${port}`)",
-            "connectCommands": "Replace all previous parameters to specify an array of commands to establish connection"
+            "connectCommands": "Array of GDB commands to establish the connection. Auto-sourced gdbinit ``source`` lines are prepended before these user-defined commands"
         }
     }
 
-An example of a modified launch.json file is shown below:
+An example of a customized ``launch.json`` is shown below. The minimal default configuration above is enough for most projects when ESP-IDF has generated ``build/gdbinit/gdbinit``. Optional fields can be added as needed:
 
 .. code-block:: JSON
 
@@ -215,23 +229,19 @@ An example of a modified launch.json file is shown below:
                 "request": "attach",
                 "name": "Eclipse CDT GDB Adapter",
                 "program": "${workspaceFolder}/build/${command:espIdf.getProjectName}.elf",
-                "initCommands": [
-                    "set remote hardware-watchpoint-limit IDF_TARGET_CPU_WATCHPOINT_NUM",
-                    "mon reset halt",
-                    "maintenance flush register-cache"
-                ],
                 "gdb": "${command:espIdf.getToolchainGdb}",
+                "initialBreakpoint": "app_main",
+                "initCommands": [
+                    "set remote hardware-watchpoint-limit 2"
+                ],
                 "target": {
                     "connectCommands": [
-                        "set remotetimeout 20",
-                        "target remote localhost:3333"
+                        "set remotetimeout 20"
                     ]
                 }
             }
         ]
     }
-
-While the previous example is explicitly using the default values, it can be customized to suit your needs.
 
 There are other, less used arguments documented in the ESP-IDF VS Code extension's package.json gdbtarget debugger contribution.
 
--- a/docs_espressif/en/debugproject.rst
+++ b/docs_espressif/en/debugproject.rst
@@ -154,25 +154,31 @@ You can modify the configuration to suit your needs. Let's describe the configur
 - ``type``: The type of the debug configuration. It should be set to ``gdbtarget``.
 - ``program``: ELF file of your project build directory to execute the debug session. You can use the command ``${command:espIdf.getProjectName}`` to query the extension to find the current build directory project name.
 - ``initCommands``: Optional GDB commands run **after** the target is connected via ``target.connectCommands``. The extension does not inject a default ``initCommands`` list.
-- ``initialBreakpoint``: When set to a non-empty string (for example ``app_main``), the extension appends ``thb <value>`` to ``initCommands``. If set to ``""`` (empty string) or omitted, no breakpoint is added from this setting.
+- ``initialBreakpoint``: When set to a non-empty string (for example ``app_main``), the extension appends ``thb <value>`` to ``initCommands``. When set to ``""`` (empty string), the session does not stop on an initial breakpoint. When omitted, the extension adds ``thbreak app_main`` to the connect sequence.
 - ``gdb``: GDB executable to be used. By default "${command:espIdf.getToolchainGdb}" will query the extension to find the ESP-IDF toolchain GDB for the current IDF_TARGET of your esp-idf project (esp32, esp32c6, etc.).
 
 .. note::
      **GDB command order.** The Eclipse CDT GDB Adapter runs ``target.connectCommands`` first to attach to the target, then runs ``initCommands``.
 
-     When files exist under ``idf.buildPath``, the extension **prepends** commands to ``target.connectCommands`` (before any user-defined ``connectCommands``), in this order:
+     The extension **prepends** commands to ``target.connectCommands`` (before any user-defined ``connectCommands``), in this order:
 
-     1. ``gdbinit/gdbinit`` — ESP-IDF aggregate script (typically sources symbols and connect; available from ESP-IDF v5.3.3). If this file is missing, the extension prepends the same connect sequence as ESP-IDF's ``gdbinit/connect``:
+     1. A ``source`` line for each ESP-IDF generated gdbinit file listed in ``gdbinit_files`` of ``build/project_description.json``, using the same order as ``idf.py gdb``: ``symbols``, ``prefix_map`` and ``py_extensions``. ``py_extensions`` is skipped when the configured GDB has no Python support. When ``gdbinit_files`` is unavailable, the extension looks for those files under ``${idf.buildPath}/gdbinit`` and falls back to ``prefix_map_gdbinit`` for path remapping.
+
+     2. The commands from ESP-IDF's ``gdbinit/connect`` file, replayed one by one instead of being sourced, with these adjustments:
+
+        - The trailing ``continue`` is dropped. Connect commands run before the debug session is initialized, so resuming the target here would run past breakpoints the editor has not inserted yet and the resulting stop would not be reported.
+        - ``thbreak app_main`` is dropped when ``initialBreakpoint`` is defined in the launch configuration, so that setting alone decides where the session first stops.
+        - ``target remote :3333`` is rewritten when ``target.host`` or ``target.port`` are defined.
+
+        When the file is missing or has a shape the extension cannot expand, such as the ``CONFIG_IDF_TARGET_LINUX`` variant that has no remote target, the extension falls back to an equivalent sequence:
 
         - ``set remotetimeout 10``
-        - ``target remote :3333``
+        - ``target remote :3333`` — uses ``target.host`` and ``target.port`` when they are defined
         - ``monitor reset halt``
         - ``maintenance flush register-cache``
-        - ``thbreak app_main``
+        - ``thbreak app_main`` — only when ``initialBreakpoint`` is omitted from the launch configuration
 
-     2. ``gdbinit/prefix_map``, or fallback ``prefix_map_gdbinit`` — path remapping for reproducible builds. This is a separate file and is **not** included inside ``gdbinit/gdbinit``.
-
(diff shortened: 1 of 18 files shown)


AndriiFilippov (approved):
@brianignacio5 hi ! 
Tested under:
OS: Windows 11 / Linux Ubuntu 24.04 / Mac arm64
ESP-IDF: v6.0.2
Boards: ESP32 Ethernet kit \ ESP32-C6.


* Generated `gdbinit/symbols` is sourced correctly and bootloader symbols are resolved ✅ 
* The original symbol-loss issue during repeated `Step Into` / `Step Over` is fixed ✅
* Reproducible build source mapping works correctly ✅ 
* Custom `initCommands` do not prevent `gdbinit/symbols` or `gdbinit/prefix_map` from being loaded  ✅
* `IDF_TARGET_CPU_WATCHPOINT_NUM` without braces is replaced correctly (`2` for ESP32, `4` for ESP32-C6) ✅

LGTM 👍

radurentea (approved):
LGTM, nice improvements!

brianignacio5:
PTAL @sobuch @gerekon for debug session stopping when OpenOCD is stopped.

sobuch:
Hi @brianignacio5 , thanks for the updates! Few notes from my side:

- Checked debug session closes after openocd shutdown (fixes https://github.com/espressif/openocd-esp32/issues/395). However there seems to be a change that closing the debug session now also shuts down openocd, which may not be desirable (PTAL @erhankur in case you have a preference).

- Updated IDF_TARGET_CPU_WATCHPOINT_NUM values are correct now, however (thinking how to avoid possible mismatches in the future), could we not retrieve this value from sdkconfig? or is that already done and this is just a backup? I am not sure here...

- We looked into the problem with symbols in detail with @AndriiFilippov , and it seems this is not completely solved here yet. The "gdbinit/symbols" file is sourced correctly, and for example "info address memset" will resolve an address from ROM. However, "info symbol 0x400011e8" (address of __call_memset, in ROM) returns "No symbol matches 0x400011e8."

- The above commands seem to work ok from commandline, and the difference appears to be that in the "build/gdbinit/connect" file we use "target remote", instead of "target extended-remote" that is used by the vscode extension. Is there a reason to favor "extended-remote", or can we also source the "gdbinit/connect" file?

- @Lapshin could you please check if the behaviour here is as expected, or should it still be possible to resolve symbols from ROM with "idf.py gdb", if I replace "target remote" -> "target extended-remote" in "build/gdbinit/connect"?

brianignacio5:
> Hi @brianignacio5 , thanks for the updates! Few notes from my side:
> 
> * Checked debug session closes after openocd shutdown (fixes [riscv32-esp-elf-gdb-no-python.exe does not close openocd-esp32#395](https://github.com/espressif/openocd-esp32/issues/395)). However there seems to be a change that closing the debug session now also shuts down openocd, which may not be desirable (PTAL @erhankur in case you have a preference).
> * Updated IDF_TARGET_CPU_WATCHPOINT_NUM values are correct now, however (thinking how to avoid possible mismatches in the future), could we not retrieve this value from sdkconfig? or is that already done and this is just a backup? I am not sure here...
> * We looked into the problem with symbols in detail with @AndriiFilippov , and it seems this is not completely solved here yet. The "gdbinit/symbols" file is sourced correctly, and for example "info address memset" will resolve an address from ROM. However, "info symbol 0x400011e8" (address of __call_memset, in ROM) returns "No symbol matches 0x400011e8."
> * The above commands seem to work ok from commandline, and the difference appears to be that in the "build/gdbinit/connect" file we use "target remote", instead of "target extended-remote" that is used by the vscode extension. Is there a reason to favor "extended-remote", or can we also source the "gdbinit/connect" file?
> * @Lapshin could you please check if the behaviour here is as expected, or should it still be possible to resolve symbols from ROM with "idf.py gdb", if I replace "target remote" -> "target extended-remote" in "build/gdbinit/connect"?

Hi @sobuch thank you for your feedback.

1. The debug session should only close OpenOCD if OpenOCD was launched when starting a debug session. If OpenOCD was started from before, it will remain running. This behavior is fixed in this PR. We could add a setting if we want the debug session to start OpenOCD but don't kill it after, I'm not sure if this is desired.
2. `IDF_TARGET_CPU_WATCHPOINT_NUM` and `target extended-remote` now the PR uses `build/gdbinit/gdbinit` if available to establish the connection to the device. If the gdbinit file doesn't exist, will use  a default value which mimics what `build/gdbinit/connect` does.

erhankur:
> Checked debug session closes after openocd shutdown (fixes https://github.com/espressif/openocd-esp32/issues/395). However there seems to be a change that closing the debug session now also shuts down openocd, which may not be desirable (PTAL @erhankur in case you have a preference).

In my opinion, if OpenOCD is guaranteed to be dedicated to a single debug session, always stopping it is reasonable. Starting a fresh OpenOCD instance avoids errors after programming, and its startup time is short.

Lapshin:
> We looked into the problem with symbols in detail with @AndriiFilippov , and it seems this is not completely solved here yet. The "gdbinit/symbols" file is sourced correctly, and for example "info address memset" will resolve an address from ROM. However, "info symbol 0x400011e8" (address of __call_memset, in ROM) returns "No symbol matches 0x400011e8.

You have to connect to target to take an effect of this handler - https://github.com/espressif/esp-idf/blob/08e0d30a74ad0bfd5a34933142b80f45619ee410/components/esp_rom/gen_gdbinit.py#L43
E.g. esp32c5 file `build/gdbinit/symbols`:
```
# Load esp32c5 ROM ELF symbols
define target hookpost-remote
set confirm off
  # if $_streq((char *) 0x4004b3c4, "Mar 29 2024")
  if (*(int*) 0x4004b3c4) == 0x2072614d && (*(int*) 0x4004b3c8) == 0x32203932 && (*(int*) 0x4004b3cc) == 0x343230
    add-symbol-file /home/alex/.espressif/tools/esp-rom-elfs/20260528/esp32c5_rev0_rom.elf
  else
    # if $_streq((char *) 0x4004aefc, "Jan 21 2025")
    if (*(int*) 0x4004aefc) == 0x206e614a && (*(int*) 0x4004af00) == 0x32203132 && (*(int*) 0x4004af04) == 0x353230
      add-symbol-file /home/alex/.espressif/tools/esp-rom-elfs/20260528/esp32c5_rev100_rom.elf
    else
      echo Warning: Unknown esp32c5 ROM revision.\n
    end
  end
set confirm on
end

....
....
```

> IDF_TARGET_CPU_WATCHPOINT_NUM and target extended-remote now the PR uses build/gdbinit/gdbinit if available to establish the connection to the device. If the gdbinit file doesn't exist, will use a default value which mimics what build/gdbinit/connect does.

No for `extended-remote` `hookpost-remote` would never be called.

Lapshin:
@brianignacio5 , in case you want to switch to `extended-remote` just change hookpost accordingly

sobuch:
Hi @brianignacio5 now I am seeing an issue that breakpoints are not getting hit after starting the debug session, neither app_main nor ones that I placed myself.

However both the ROM symbol resolution and OpenOCD/GDB closing seem to work consistently and as expected :+1:

brianignacio5:
Thanks @Lapshin ! I've tested and we can just switch to `target remote` to avoid changing the hook post.

@sobuch This happens because the `build/gdbinit/connect` has a `continue` after setting breakpoint to `app_main`. I've tested removing the continue line in `build/gdbinit/connect` and the `app_main` stops as it was doing before. 

To solve this issue we either 

1. Update generated gdbinit file from ESP-IDF.
2. Go back to defining in extension those `build/gdbinit/connect` lines without `continue`
3. Always patch `build/gdbinit/connect` to remove `continue` line.
4. Another solution you guys think ?

Easiest I can do is 2 in this PR. What do you think @sobuch ?

sobuch:
> @sobuch This happens because the `build/gdbinit/connect` has a `continue` after setting breakpoint to `app_main`. I've tested removing the continue line in `build/gdbinit/connect` and the `app_main` stops as it was doing before.

Indeed, I can see the same. In logs from openocd its also visible that the breakpoint does get hit, however another request to continue is sent right after. Do you know where that is coming from? I only see continues in "setDataBreakpointsRequest"

brianignacio5:
> > @sobuch This happens because the `build/gdbinit/connect` has a `continue` after setting breakpoint to `app_main`. I've tested removing the continue line in `build/gdbinit/connect` and the `app_main` stops as it was doing before.
> 
> Indeed, I can see the same. In logs from openocd its also visible that the breakpoint does get hit, however another request to continue is sent right after. Do you know where that is coming from? I only see continues in "setDataBreakpointsRequest"

From the `<esp-idf-project>/build/gdbinit/connect`. This is not a vscode request, it is coming from the gdbinit file itself. you can see after the `source <esp-idf-project>/build/gdbinit/connect`

sobuch:
> From the `<esp-idf-project>/build/gdbinit/connect`. This is not a vscode request, it is coming from the gdbinit file itself. you can see after the `source <esp-idf-project>/build/gdbinit/connect`

That doesnt seem correct, if you remove this `continue` form `gdbinit/connect` you get breakpoint hit in `app_main` as well, right? So another request to continue must be coming from somewhere else, as after the connect script the target would just be halted at the reset vector (due to `monitor reset halt`).

brianignacio5:
The problem is that the connect gdbinit file is executed in the launchRequest or attachRequest which is executed after the setBreakpoints request.

See Launch Sequencing in https://microsoft.github.io/debug-adapter-protocol/overview

connect 
```
set remotetimeout 10
target remote :3333
monitor reset halt
maintenance flush register-cache
thbreak app_main
continue
```

sobuch:
> The problem is that the connect gdbinit file is executed in the launchRequest or attachRequest which is executed after the setBreakpoints request.

Well what happens is gdbinit leaves the target halted at the breakpoint and then it is resumed again. I was mainly wondering where thats coming from, found it here - https://github.com/espressif/vscode-esp-idf-extension/blob/857e663907b43c0823c1cbd82f885b593b6aaee3/src/cdtDebugAdapter/adapter/GDBDebugSession.ts#L960  

I think theres some value in using the same gdbinit as from commandline, so I would suggest adding a flag whether to resume in the `configurationDoneRequest`. That could be false by default if the gdbinit file is used and true otherwise, maybe with an extra option for launch config for users to override it.

brianignacio5:
> > The problem is that the connect gdbinit file is executed in the launchRequest or attachRequest which is executed after the setBreakpoints request.
> 
> Well what happens is gdbinit leaves the target halted at the breakpoint and then it is resumed again. I was mainly wondering where thats coming from, found it here -
> 
> https://github.com/espressif/vscode-esp-idf-extension/blob/857e663907b43c0823c1cbd82f885b593b6aaee3/src/cdtDebugAdapter/adapter/GDBDebugSession.ts#L960
> 
> I think theres some value in using the same gdbinit as from commandline, so I would suggest adding a flag whether to resume in the `configurationDoneRequest`. That could be false by default if the gdbinit file is used and true otherwise, maybe with an extra option for launch config for users to override it.

I'm not sure is related. 

I've tested by commenting `await mi.sendExecContinue(this.gdb);`. If I just want to hit the breakpoint after app_main, for example I use the blink example with esp32 wrover kit and set a breakpoint in `blink_led()` line, commenting or keeping `await mi.sendExecContinue(this.gdb);` doesn't matter, I still reach the line after app_main breakpoint (the app_main bp only stop if I remove continue from gdbinit file).

brianignacio5:
PTAL again @sobuch The breakpoint issue should be fixed with latest commit.

sobuch:
Hi @brianignacio5 for me it now seems to stop on breaks that I place myself, but not on the app_main from gdbinit, do you see the same?

In the openocd -d3 logs I can see the break was placed, just it again resumed after hit:

```
Info : 1425 85744 esp_xtensa.c:269 esp_xtensa_reset_reason_read(): [esp32s3.cpu0] Reset cause (3) - (Software core reset)
... add breakpoint:
Debug: 1439 85749 gdb_server.c:408 gdb_log_incoming_packet(): [esp32s3.cpu0] {4} received packet: Z1,42008e23,2
... resume:
Debug: 1446 85749 gdb_server.c:408 gdb_log_incoming_packet(): [esp32s3.cpu0] {4} received packet: vCont;c
... hit breakpoint/target halted:
Debug: 1965 86142 gdb_server.c:426 gdb_log_outgoing_packet(): [esp32s3.cpu0] {4} sending packet: $T05thread:3fc9674c;#df
... remove breakpoint:
Debug: 1978 86144 gdb_server.c:408 gdb_log_incoming_packet(): [esp32s3.cpu0] {4} received packet: z1,42008e23,2
... resume:
Debug: 1985 86152 gdb_server.c:408 gdb_log_incoming_packet(): [esp32s3.cpu0] {4} received packet: vCont;c
```

brianignacio5:
Yes but the breakpoint in app_main from gdbinit file will be ignored because of the `continue` after it. 

As I said before, even if I remove `await mi.sendExecContinue(this.gdb);` from debug adapter, the app_main breakpoint would not stop for the user. For me, seems that the issue comes from `continue` from gdbinit.

Lapshin:
@brianignacio5 , let's take a look at the generagted connect file:

```
# Connect to the default openocd-esp port and stop on app_main()
set remotetimeout 10
target remote :3333
monitor reset halt
maintenance flush register-cache
thbreak app_main
continue
```

After `target remote :3333` the next command is `monitor reset halt` which **resets** the device and **stops** execution.

If `continue` is removed, how will it be run again?
It looks like you have unexpected `continue` somewhere in scripts

brianignacio5:
> @brianignacio5 , let's take a look at the generagted connect file:
> 
> ```
> # Connect to the default openocd-esp port and stop on app_main()
> set remotetimeout 10
> target remote :3333
> monitor reset halt
> maintenance flush register-cache
> thbreak app_main
> continue
> ```
> 
> After `target remote :3333` the next command is `monitor reset halt` which **resets** the device and **stops** execution.
> 
> If `continue` is removed, how will it be run again? It looks like you have unexpected `continue` somewhere in scripts

I understand. The issue is the Debug adapter implementation and how the connection request is organized. In summary, the issue is that the debug adapter is not initialized yet when the breakpoint is hit, so the UI doesn't capture the breakpoint even though in CLI is hit. A major refactor would be needed so I just made some adjustments in the mean time.

In the future I think we can explore using the GDB DAP directly and just have configuration to get ESP-IDF build files as idf.py gdb does.

sobuch:
yes, thanks for the updates @brianignacio5 , I tested the latest version and worked ok for me.

One unrelated note - it could be nice to print a warning for user to maybe remove `"runOpenOCD": false` when its set and the connection fails, in case _someone_ just forgets about the setting

PR #1085, opened 2023-11-24
Fix headers and include resolution by specifying compileCommand.json path

radurentea:
## Description

Added compileCommands.json path, this will help getting rid of the squiggly lines in C files for the esp projects after build.

Task: https://jira.espressif.com:8443/browse/VSC-1234

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- This change requires a documentation update

## Steps to test this pull request

Provide a list of steps to test changes in this PR and required output

1. Create new esp project
2. Build project
3. Open the main c file of the project (Ex for blink project: blink_example_main.c)

- Expected behaviour:
The file should display no errors (no squiggly red lines)

## How has this been tested?

Created esp projects using all possible options in VS Code Extension on Windows 11 with no issues shown after building the projects.

**Test Configuration**:
* ESP-IDF Version: 5.1.2
* OS (Windows,Linux and macOS): Windows 11

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [x] Added Documentation
- [ ] Added Unit Test
- [ ] Verified on all platforms - Windows,Linux and macOS

PR #1946, opened 2026-09-15
Bugfix/vscode folder multiple configuration

### LINKED ISSUE #1872: Generated c_cpp_properties.json breaks IntelliSense compiler probing for built-in headers (e.g. stddef.h) (VSC-1958)
### OS

Linux

### Operating System version

Fedora 44

1.124.2

### ESP-IDF version

6.0.1

### Python version

3.13.5

### Doctor command output

N/A

### Extension

_No response_

### Description

When generating a new project using the ESP-IDF extension, the automatically generated .vscode/c_cpp_properties.json breaks the Microsoft C/C++ extension's compiler probing mechanism. As a result, standard built-in compiler headers (such as <stddef.h>) cannot be resolved, resulting in red squiggly lines (errors) out of the box, even though the project compiles successfully.

### Debug Message

N/A

### Other Steps to Reproduce

I found that selecting "Do not configure with a compiler (not recommended)" in C/C++: Select IntelliSense Configuration... (Ctrl + Shift + P) solves the issue and all includes are correctly resolved. Clicking this button edits the c_cpp_properties.json file in this way:

{
    "configurations": [
        {
            "name": "ESP-IDF",
            "compilerPath": "",
            "compileCommands": [
                "${config:idf.buildPath}/compile_commands.json"
            ],
            "intelliSenseMode": "gcc-x86",
            "includePath": [
                "${workspaceFolder}/**"
            ],
            "browse": {
                "path": [
                    "${workspaceFolder}"
                ],
                "limitSymbolsToIncludedHeaders": true
            }
        }
    ],
    "version": 4
}

so maybe if the extension can produce a c_cpp_properties.json like this one, with compilerPath left empty, the issue can be solved

### LINKED ISSUE #1881: Add VS Code Configuration Folder doesn't add correct compileCommands with multiple configurations (VSC-1966)
### OS

Linux

### Operating System version

Ubuntu 24.04.4 LTS

1.125.1

### ESP-IDF version

6.0.1

### Python version

3.12.3

### Doctor command output

[report.txt](https://github.com/user-attachments/files/29166842/report.txt)

### Extension

[esp_idf_vsc_ext.log](https://github.com/user-attachments/files/29166848/esp_idf_vsc_ext.log)

### Description

When I select my `test_refresh` configuration, the `c_cpp_properties.json` file is updated to contain the following:

```
"compileCommands": "/home/jdbaptista/Documents/repos/traffic-pcb-sb/software/esp-firmware/builds/test_refresh/compile_commands.json",
```

When I am already in the `test_refresh` configuration, delete the .vscode folder, then run the `Add VS Code Configuration Folder` command, the file instead contains:

```
"compileCommands": "${config:idf.buildPath}/compile_commands.json",
```

I am not aware of any issues that may arise from this, but I figure they should match. If this is fine, then I will ignore it. The only issue is that VSCode presents the following warning due to idf.buildPath not being the correct folder (build rather than builds).

```
Cannot find: "/home/jdbaptista/Documents/repos/traffic-pcb-sb/software/esp-firmware/build/compile_commands.json"
```

### Debug Message

none.

### Other Steps to Reproduce

_No response_

radurentea:
Fix two problems with the `.vscode/c_cpp_properties.json` the extension generates, and make `ESP-IDF: Add VS Code Configuration Folder` configure clangd the same way project creation does.

**Preset build path (#1881).** The template points `compileCommands` at `${config:idf.buildPath}/compile_commands.json`, which the C/C++ extension expands from `settings.json`. A selected configure preset overrides `idf.buildPath` in memory only, so after regenerating the folder the file pointed at the default `build` directory and the C/C++ extension reported `Cannot find: .../build/compile_commands.json`. Selecting the preset produced the correct file because preset selection re-applies the resolved path.

**Compiler override (#1872).** The template shipped `"compilerPath": "${default}"` and `createVscodeFolder` replaced it with the absolute path of the toolchain gcc found on the PATH. A non-empty `compilerPath` makes the C/C++ extension query that compiler for system includes and defines instead of the one in `compile_commands.json`. When the toolchain was not on the PATH at creation time the placeholder stayed and the extension queried a host compiler or nothing, so built-in headers such as `stddef.h` could not be resolved. When the lookup succeeded, the file carried a machine-specific path for the window's current target rather than the project's, which went stale on toolchain updates and target changes. `${default}` dates from 2020 (#195), before the template had `compileCommands` at all; since #1085 added `compileCommands` the override only matters when it is wrong.

**Add VS Code Configuration Folder:**

* New `addVscodeFolderToWorkspace` in `src/newProject/utils.ts` holds the command's sequence: copy the templates, resolve `compileCommands` through `idf.buildPath` (which honours a selected preset or the workspace setting), and configure clangd settings. The command handler makes that single call.
* `createVscodeFolder` stays the plain template copy used by the new-project flows, so a new project never inherits the preset selected in the window it was created from.
* The `idf.buildPath` change handler now refreshes `c_cpp_properties.json` alongside the clangd arguments, since the file carries a resolved path.

**compilerPath:**

* The template ships `"compilerPath": ""` and nothing rewrites it. The C/C++ extension takes the compiler from each `compile_commands.json` entry, which is what "Do not configure with a compiler" writes and what fixed the report.
* `ESP-IDF: Set Espressif Device Target` replaces its old compiler-path rewrite with `clearCCppPropertiesJsonCompilerPath`, which empties a non-empty `compilerPath` so existing projects do not keep a stale, wrong-architecture override after a target change. Missing, unparsable, or already-empty files are not touched.
* Trade-off: with an empty `compilerPath` there is no compiler query for files that have no `compile_commands.json` entry. That covers everything before the first build, where IntelliSense was already unusable because the ESP-IDF include paths only exist after a build, and files outside the build such as an excluded component. The extension already prompts to generate `compile_commands.json` when it is missing.

**Tests (`src/test/project.test.ts`):**

* Generated `c_cpp_properties.json` matches the template and keeps an empty `compilerPath`.
* `createVscodeFolder` alone keeps the template `compileCommands`; `addVscodeFolderToWorkspace` writes the absolute preset path with a preset selected and the resolved default `build` path without one.
* clangd arguments target the preset build directory (seeds the IDF environment from the test host; skips when esp-clangd is not on the PATH).
* `clearCCppPropertiesJsonCompilerPath` empties an absolute path, leaves an empty one untouched without rewriting the file, and ignores a missing file.

**Docs:** `configureproject.rst` (en, zh_CN) now shows the shipped template and explains the empty `compilerPath` and when `compileCommands` is rewritten. CHANGELOG entries added.

Not covered here: the selected preset is stored per VS Code window, so `readParameter` applies it to any scope, including a new project being created from that window. That is a separate bug with its own report.

Closes #1881
Closes #1872

## Type of change

1. Open an ESP-IDF project with a `CMakePresets.json` that defines a configure preset with a non-default `binaryDir`, for example `builds/test_refresh`.
2. Run `ESP-IDF: Select Project Configuration` and pick that preset.
3. Delete the `.vscode` folder from the project.
4. Run `ESP-IDF: Add VS Code Configuration Folder`.
5. Open `.vscode/c_cpp_properties.json` and check `compilerPath` and `compileCommands`.
6. With a standard ESP-IDF tools install (esp-clangd present), open `.vscode/settings.json` and check `clangd.path` and `clangd.arguments`.
7. Repeat steps 3 and 4 with no preset selected and check both files again.
8. In a project whose `c_cpp_properties.json` still has an absolute `compilerPath`, run `ESP-IDF: Set Espressif Device Target` and check the file.
9. Build the project, open a source file that includes `stddef.h`, and confirm the C/C++ extension reports no missing built-in headers.

- Expected behaviour:

`compilerPath` is `""` in every generated file. With a preset selected, `compileCommands` is the absolute path to `compile_commands.json` under the preset `binaryDir`, and `clangd.arguments` contains `--compile-commands-dir=<preset build directory>`. Without a preset, both point at the project's `build` directory. Set Target empties an existing absolute `compilerPath` and leaves the rest of the file alone. After a build, built-in headers resolve and the `Cannot find` warning is gone.

- Expected output:

    "compilerPath": "",
    "compileCommands": "<project>/builds/test_refresh/compile_commands.json"
    "--compile-commands-dir=<project>/builds/test_refresh"

## How has this been tested?

Manual test with the steps above. Ran `yarn test`: the new project tests pass and the remaining failures are identical on `master` and specific to the local Windows environment.

**Test Configuration**:
* ESP-IDF Version: 6.1

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting
- [x] Added Documentation
- [x] Added Unit Test


src/newProject/utils.ts
```diff
@@ -119,6 +123,44 @@ export async function createVscodeFolder(
...
+ * createVscodeFolder directly so they never inherit another folder's preset.
+ * @param {string} extensionPath - Extension root that holds the templates.
+ * @param {Uri} workspaceFolder - Workspace folder to add the files to.
+ */
+export async function addVscodeFolderToWorkspace(
+  extensionPath: string,
+  workspaceFolder: Uri
+) {
+  await createVscodeFolder(extensionPath, workspaceFolder);
+  await applySelectedProjectConfigurationToVscodeFolder(workspaceFolder);
```
brianignacio5, line 139: I think this function `applySelectedProjectConfigurationToVscodeFolder` is not necessary.

We can replace this line with `await setCCppPropertiesJsonCompileCommands(workspaceFolder);`

The function itself would be better, because it will always resolve the build path and there is no need to check a selectedConfig exists, because we always want the JsonCompileCommands to have the correct build path.

brianignacio5 (changes requested):
In my opinion, there is an unnecessary redundancy that is not needed.

src/newProject/utils.ts
radurentea: Thanks, I agree.
I've replaced it with the direct call. However, there is one side effect: without a preset the file now gets the resolved default build path instead of `${config:idf.buildPath}`, so I also made the `idf.buildPath` change handler refresh `c_cpp_properties.json` the way it already refreshes the clangd arguments. Otherwise the path would go stale when the setting changes.

The commit covering the change is [579e441](https://github.com/espressif/vscode-esp-idf-extension/pull/1946/commits/579e441edb07f889540a97589106826580ad1ca2)

brianignacio5 (approved):
LGTM

src/newProject/index.ts
```diff
@@ -140,6 +142,8 @@ async function runCreateVsCodeFolder(
 ) {
   try {
     await createVscodeFolder(context.extensionPath, wsFolder.uri);
+    await applySelectedProjectConfigurationToVscodeFolder(wsFolder.uri);
+    await configureClangSettings(wsFolder.uri);
```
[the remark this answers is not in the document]
radurentea: **showError: true** also throws **idfToolNotFound** when esp-clangd is not on the PATH, which would make the command fail on every machine without it. This call intentionally matches project creation and preset selection, which pass **false**. An invalid **settings.json** is logged by **configureClangSettings**; surfacing that separately from the tool-not-found case would be a change to **configureClangSettings** itself and is out of scope here.

AndriiFilippov:
@radurentea hi ! 
Tested under: 
OS: Windows 11

**Steps to reproduce:**

1. Open a multi-config project.
2. Select a configuration, for example `test_second`, with build directory:
   `...\multi_config_R_test\builds\test_second`
3. Create a new project from `examples/blink`.
4. Open the new project's `.vscode/settings.json`.

Actual result:

```json
"clangd.arguments": [
  "--background-index",
  "--query-driver=**",
  **"--compile-commands-dir=c:\\Users\\andri\\multi_config_R_test/builds/test_second"**
]
```

The new project incorrectly inherits the clangd compile commands directory from the previously opened multi-config project.

At the same time, `.vscode/c_cpp_properties.json` is generated correctly:

```json
"compilerPath": "",
"compileCommands": "${config:idf.buildPath}/compile_commands.json"
```

Expected result:

The new project should use its own build directory, for example:

```json
"--compile-commands-dir=c:\\Users\\andri\\blinkR_simple_project\\build"
```

or another value resolved from the new project's own `idf.buildPath`, but it should not reference the previous project's build directory.

So it looks like the new-project flow correctly resets `c_cpp_properties.json`, but `clangd.arguments` still keeps the previous project's selected build configuration.
[doctor.txt](https://github.com/user-attachments/files/32456454/doctor.txt)

@radurentea  @brianignacio5 
Please make sure to assign at least 2 reviewers for PRs going forward, as required by our review process.
Thanks !

radurentea:
Hi @AndriiFilippov,

Thanks for testing. Confirmed, and it is a separate pre-existing bug rather than a regression from this PR: the selected preset is stored once per window and readParameter applies it to any scope, so setClangSettings during project creation picks up the previous project's resolved build directory.

I will create a PR soon with a fix.

